PR #1, opened 2016-02-16
Update README.md

Yoric:


README.md
```diff
@@ -1,6 +1,6 @@
 # ![Foxbox Client](https://raw.githubusercontent.com/fxbox/app/master/app/img/icons/32.png "Foxbox Client") Foxbox Client
 
-> A client for Foxbox
+> A web client for Foxbox (with support for emulating the server component in the meantime).
```
michielbdejong: Which parts of the server component does it emulate? You mean the fact that the `gulp` command starts a web server with this app's static content on `http://localhost:8000/`, or is there more?

Yoric: Exactly that.

michielbdejong: Right, in that case I think it might be confusing to state that explicitly, since it's not really that our app support it, more that `gulp` does, and all client-side javascript build systems offer that functionality (not only `gulp` but also `grunt` and `bower` IIRC).

Also, as we're about to add a Cordova build target, we could say something "A web client for Foxbox" and leave out the remark about `gulp`'s static webserver support.

PR #3, opened 2016-02-23
Add Cordova build target

michielbdejong:
Looks like `gulp` is building ES6 - should find out how to generate ES5 code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,31 @@ $ gulp
 Then point your browser to [http://localhost:8000](http://localhost:8000/).
 
 Note: The site is built in the `dist/app` folder.
+
+## Building for Cordova
+
+To build this app as a Cordova app, run:
+
+```bash
+npm install -g cordova
+gulp cordova-setup
+gulp cordova-android
+```
+
+Make sure you have either an Android device connected over the Android Debugging
+Bridge, or have the Android emulator installed.
+
+If the Android emulator starts but the app doesn't open, you can hit `^C` and
+run `cd dist/cordova ; cordova emulate android` again.
+
+You can use Chrome's dev tools -> more tools -> inspect devices
+to debug the Android emulator.
+
+For local development of the SecureHTTP plugin, you can do:
+
+```bash
+cd dist/cordova/
+cordova plugin remove com.synconset.cordovaHTTP
+cordova plugin add /path/to/local/SecureHTTP
+cordova run android
+```
--- a/app/cordova.js
+++ b/app/cordova.js
@@ -1,0 +1,1 @@
+// To be generated during Cordova build.
--- a/app/index.html
+++ b/app/index.html
@@ -16,6 +16,7 @@
 <link rel="icon" href="img/icons/32.png" sizes="32x32">
 <link rel="icon" href="img/icons/512.png" sizes="512x512">
 <title>Foxbox Client</title>
+<script defer src="cordova.js"></script>
 <script defer src="components/react/react.js"></script>
 <script defer src="components/react/react-dom.js"></script>
 <script defer src="js/initapp.js"></script>
--- a/app/components/fxos-mvc/dist/mvc.js
+++ b/app/components/fxos-mvc/dist/mvc.js
@@ -1320,8 +1320,13 @@
     /**
      * Escapes HTML for all values in a tagged template string.
      */
-    escapeHTML: function (strings, ...values) {
+    escapeHTML: function (strings) {
       var result = '';
+      var _len = arguments.length;
+      var values = new Array(_len > 1 ? _len - 1 : 0);
+      for (var _key = 1; _key < _len; _key++) {
+        values[_key - 1] = arguments[_key];
+      }
 
       for (var i = 0; i < strings.length; i++) {
         result += strings[i];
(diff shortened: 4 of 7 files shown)


michielbdejong:
r? @gmarty

Thanks for looking into the ES5 issue, that worked! :)

app/index.html
```diff
@@ -16,6 +16,7 @@
 <link rel="icon" sizes="512x512" href="img/icons/512.png">
 <meta name="application-name" content="Foxbox Client">
 <title>Foxbox Client</title>
+<script type="text/javascript" src="cordova.js"></script>
```
michielbdejong: This is needed for the Cordova build; I added an empty `app/cordova.js` file to the repo so that it won't error when using `gulp` to serve the app on http://localhost:8000/. We should do this in a fancier way eventually, but I think for now it does the trick. :)

gmarty:
This PR looks good to me, but I think we should wait until the es5 issue is fixed in the MVC repo then we can just update the dependencies. This issue is being tracked at https://github.com/fxos-eng/sanitizer/pull/3. Let's see how it goes.

michielbdejong:
OK! Will merge it without the `app/components/fxos-mvc/dist/mvc.js` changes once that's updated in master.

app/cordova.js
```diff
@@ -0,0 +1 @@
+<!-- to be generated during Cordova build -->
```
gmarty: Can you make it a regular comment? It seems that [HTML style comments are not supported by Microsoft browsers](https://kangax.github.io/compat-table/es6/#test-HTML-style_comments).

michielbdejong: Ah of course, this is a .js file, wrong comment style.

README.md
```diff
@@ -30,3 +30,50 @@ $ gulp
...
+cordova create cordova
+rm -r cordova/www/*
+cp -r app/* cordova/www/
+rm cordova/www/cordova.js
+cd cordova
+cordova plugin add https://github.com/michielbdejong/SecureHTTP.git
+cordova platform add android
+cordova build android
+cordova emulate android
+```
```
gmarty: Do you think you can make this a gulp task in the `gulpfile.js` instead?
The command to build would just have to be `gulp cordova`.

app/index.html
gmarty: Can you remove the `type="text/javascript"` attribute and add a `defer` one, for the sake of consistency?

README.md
michielbdejong: ok, will have a look at that!

app/cordova.js
michielbdejong: Done

README.md
michielbdejong: Done. I couldn't merge `gulp cordova-setup` and `gulp cordova-android` into one task because the second has to run in a folder (`dist/cordova`), which the first one creates. So I left it as two tasks, hope that's ok.

app/index.html
michielbdejong: Done

michielbdejong:
> This PR looks good to me, but I think we should wait until the es5 issue is fixed in the MVC repo then we can just update the dependencies. This issue is being tracked at fxos-eng/sanitizer#3. Let's see how it goes.

Can we just merge it now and then iterate later?

app/js/lib/foxbox-qr.js
```diff
@@ -0,0 +1,50 @@
...
+const FINGERPRINT_PART = 1;
+
+export default class Qr {
+  constructor() {
+    if (cordova && cordova.plugins &&
+        cordova.plugins.barcodeScanner &&
+        cordovaHTTP && cordovaHTTP.acceptHss) {
+      this.supported = true;
+    } else {
+      this.supported = true;
```
gmarty: Shouldn't this line read?
```javascript
this.supported = false;
```

app/js/lib/foxbox-qr.js
```diff
@@ -0,0 +1,50 @@
...
+      console.warn('QR code based discovery disabled (Cordova plugins not available)');
```
gmarty: I'm not fan of a warning that'll show all the time in the console even if running from a browser.
Can you return earlier from the constructor if we're not in a Cordova app and then show the warning only if in a Cordova app and something went wrong?

app/js/lib/foxbox.js
```diff
@@ -63,6 +64,7 @@ export default class Foxbox extends Service {
     super();
     this.settings = new FoxboxSettings();
     this.db = new FoxboxDb();
+    this.qr = new FoxboxQr();
```
gmarty: I'm really not sure about adding another dependency to the foxbox.js lib.
My plan for the future is to make this lib an independent project, whose goal is to ease communication with the box by abstracting away authentication and API calls.
What you're achieving here shouldn't be taken care of by this script IMO.

app/js/lib/foxbox-qr.js
michielbdejong: Thanks! Fixed.

app/js/lib/foxbox-qr.js
michielbdejong: OK, moved the warning out of the constructor.

app/js/lib/foxbox.js
michielbdejong: OK, moved `Qr` out of `Foxbox`.

michielbdejong:
@gmarty addressed your latest review comments.

PR #8, opened 2016-03-04
Adding discovery mechanism to the app r=gmarty

arcturus:


app/js/lib/foxbox-settings.js
```diff
@@ -12,7 +13,8 @@ export default class FoxboxSettings extends Model {
       _scheme: localStorage.getItem('foxbox-scheme') || DEFAULT_SCHEME,
       _hostname: localStorage.getItem('foxbox-hostname') || DEFAULT_HOSTNAME,
       _port: localStorage.getItem('foxbox-port') || DEFAULT_PORT,
-      _session: localStorage.getItem('session')
+      _session: localStorage.getItem('session'),
+      _skipDiscovery: localStorage.getItem('foxbox-skipDiscovery') === 'true' || false
```
gmarty: nit: Actually we don't need the `|| false` here.
Even if the entry is not defined in the local storage, we still get false (undefined !== 'true').

app/js/lib/foxbox-settings.js
```diff
@@ -59,4 +61,17 @@ export default class FoxboxSettings extends Model {
...
+  get registrationService() {
+    return REGISTRATION_SERVICE;
+  }
+
+  get skipDiscovery() {
+    return this._skipDiscovery;
+  }
+
+  set skipDiscovery(value) {
+    this._skipDiscovery = !!value;
```
gmarty: Maybe do `value = !!value;` before these lines to do the coercion only once.

PR #11, opened 2016-03-09
Create first (simple) end-to-end and integration tests

JohanLorenzo:
WIP

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,3 +58,17 @@ cordova plugin remove com.synconset.cordovaHTTP
 cordova plugin add /path/to/local/SecureHTTP
 cordova run android
 ```
+
+## Tests
+
+### All regular tests
+
+Run `npm test`.
+
+### Run end-to-end tests (e2e)
+
+Steps (to be scripted soon):
+1. Clone [foxbox](https://github.com/fxbox/foxbox/)
+2. Run it with `cargo run`
+3. With a browser, perform the first time setup by going to [http://localhost:3000](http://localhost:3000/)
+4. Run `npm test-e2e`
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+sudo: false
+language: node_js
+node_js:
+  - "4.2"
+cache:
+  directories:
+    - node_modules
+
+addons:
+  firefox: latest # Needed to support react
+
+install:
+  - "sh -e /etc/init.d/xvfb start"
+  - "export DISPLAY=:99.0"
+  - "wget http://selenium-release.storage.googleapis.com/2.50/selenium-server-standalone-2.50.1.jar"
+  - "java -jar selenium-server-standalone-2.50.1.jar > /dev/null &"
+  - npm install
+
+script:
+  - npm test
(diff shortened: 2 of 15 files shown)


tests/e2e/lib/views/login/accessors.js
```diff
@@ -0,0 +1,24 @@
...
+  get usernameField() {
+    return this.waitForElement(By.css('input[type="text"]'));
+  },
+
+  get passwordField() {
+    return this.waitForElement(By.css('input[type="password"]'));
+  },
+
+  get submitButton() {
+    return this.waitForElement(By.css('button'));
```
JohanLorenzo: We could probably have a better selector for it. What do you think @gmarty ?

tests/e2e/lib/views/login/accessors.js
```diff
@@ -0,0 +1,24 @@
...
+var _ = require('lodash');
+
+
+function LoginAccessors() {
+  Accessors.apply(this, arguments);
+}
+
+LoginAccessors.prototype = _.assign({
+  get usernameField() {
+    return this.waitForElement(By.css('input[type="text"]'));
```
JohanLorenzo: We could probably have a better selector for it. What do you think @gmarty ?

JohanLorenzo:
Here's the first revision of an E2E test. There are a couple of things that have to be addressed: 
* we require some manual set up to make them run
* No CI runs the tests, yet (because of the previous point). As a consequence, the test might bitrot.

I'd propose to handle these issues in a follow up, for 2 main reasons:
* Executing E2E tests on travis will require a fair amount of effort in set up (clone repos, make builds, run the server and go through the set up app). This'll make the PR heavier to review
* The login test can be executed against a fake foxbox In this case we require to create that fake server. Also some heavy work to review.

I'd suggest to first deal with the fake server, which looks simpler to deal with. Especially, since we have something equivalent in fxbox/foxbox#176.

r? @gmarty 

cc @isabelrios @npark-mozilla

deploy.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/bash
 # Deploys to GitHub Pages.
-# Make sure to stash your local changes before running this script.
```
gmarty: There's a npm postinstall script that revert some files. I need to fix that but in the meantime, please revert deploy.sh and gulpfile.js or don't commit the changes.

gmarty:
The login mechanism is currently being changed. It's now redirect to the box to get a token back.
Can you please wait until the new auth lands before merging this branch?

JohanLorenzo:
Updated to use the new login flow. Test is now passing on Travis. r? @gmarty

README.md
```diff
@@ -58,3 +58,13 @@ cordova plugin remove com.synconset.cordovaHTTP
 cordova plugin add /path/to/local/SecureHTTP
 cordova run android
 ```
+## Tests
+### All regular tests
+Run `npm test`.
+
+### Run end-to-end tests (e2e)
```
gmarty: Please breaklines before and after headers.

JohanLorenzo: Done

gmarty:
Can you squash all your commits into a single one?
The rest looks good to me!

JohanLorenzo:
Commits squashed and build still green. Merging.

PR #12, opened 2016-03-11
Connect to registration server over https, r=arcturus

michielbdejong:


michielbdejong:
I enabled https://knilxof.org:4243/ping with a nodejs-proxy for now (see `pm2 list` / `pm2 logs proxy` while logged in as 'ubuntu'), we can iterate on that.

arcturus:
Thx!

PR #13, opened 2016-03-15
Add IP Camera view. r=gmarty

azasypkin:
PR still depends on several foxbox and users repo PRs, but I want to start discussion early :)


app/css/vendors/ip-camera.css
```diff
@@ -0,0 +1,77 @@
+.camera-view {
```
azasypkin: __question:__ is ```vendor``` folder for service specific styles or it's more about _real_ vendors/producers?

app/css/vendors/ip-camera.css
```diff
@@ -0,0 +1,77 @@
...
+  display: flex;
+  flex-direction: column;
+}
+
+.camera-view__body {
```
azasypkin: __question:__ Do you have already any guidelines for the CSS naming/style? I was trying something BEM-like to have less nested selectors. What do you think?

app/css/vendors/ip-camera.css
```diff
@@ -0,0 +1,77 @@
...
+  flex: 1;
+}
+
+/** HACK until we unify markup for all views, reset footer absolute positioning **/
+.camera-view .footer {
```
azasypkin: Footer is currently absolutely positioned, it's just temporal hack so that I can manage view space as flexible as I need (basically I needed middle content to occupy all the space between header and footer). Or maybe I've just missed the original idea behind footer..... Tell me what you think!

app/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
 <meta charset="utf-8">
-<meta http-equiv="Content-Security-Policy" content="default-src 'self' ; style-src 'self' 'unsafe-inline' ; connect-src * ; object-src 'none' ; referrer no-referrer ;">
+<meta http-equiv="Content-Security-Policy" content="default-src 'self' ; style-src 'self' 'unsafe-inline' ; connect-src * ; object-src 'none' ; img-src * ; referrer no-referrer ;">
```
azasypkin: __question:__ I need to investigate this further, what I need is image-src that points to foxbox (foxbox provides endpoint that renders jpeg directly). What restrictions\rules do we have here?

app/js/controllers/service.js
```diff
@@ -3,12 +3,27 @@
 import { Controller } from 'components/fxos-mvc/dist/mvc';
 
 import Service from 'js/views/service';
+import CameraService from 'js/views/camera-service';
 
 export default class ServiceController extends Controller {
   main(id) {
```
azasypkin: __question:__ Here is a big question :) What was your plan for choosing appropriate view for specific devices? At what stage should we choose what view we'd like to render? For now I'm requesting service here to know what view should I render and pass service to the view itself.

app/js/lib/foxbox.js
```diff
@@ -10,6 +10,8 @@ import FoxboxDb from './foxbox-db';
 const settings = new FoxboxSettings();
 const db = new FoxboxDb();
 
+const QUERY_STRING_AUTH_TOKEN_NAME = 'auth';
```
azasypkin: This is dependency to https://github.com/fxbox/users/pull/69

app/js/lib/foxbox.js
```diff
@@ -295,15 +297,12 @@ export default class Foxbox extends Service {
    * @return {Promise}
    */
   getServiceState(id) {
-    return new Promise((resolve, reject) => {
-      fetchJSON(`${this.origin}/services/${id}/state`)
-        .then(res => {
-          if (!res) {
-            return reject(new Error(`The action couldn't be performed.`));
-          }
+    return fetchJSON(`${this.origin}/services/${id}/state`).then(res => {
```
azasypkin: __question:__ what is the rationale behind additional wrapping ```Promise```? I see it in some other places as well.

app/js/lib/foxbox.js
```diff
@@ -327,6 +326,25 @@ export default class Foxbox extends Service {
     });
   }
 
+  performServiceCommand(id, command, method = 'GET', state = undefined) {
```
azasypkin: __question:__ What is your plan to perform various service-specific commands? For now I've exposed generic ```performServiceCommand```. What do you think?

app/js/lib/foxbox.js
```diff
@@ -327,6 +326,25 @@ export default class Foxbox extends Service {
...
+    return fetchJSON(`${this.origin}/services/${id}/${command}`, method, state);
+  }
+
+  getSecureURL(sourceURL) {
```
azasypkin: Naming is hard.... But we need something like this to add ```auth``` token to the resource strings (like I mentioned before jpeg link pointing to the box, that should return JPEG directly).

app/js/views/camera-service.jsx
```diff
@@ -0,0 +1,90 @@
+/* global React */
```
azasypkin: Here is example... depends on several foxbox PRs. I'll wrap up something very soon so that you can test this.

app/js/views/services-list-item.jsx
```diff
@@ -38,19 +38,12 @@ export default class ServicesListItem extends React.Component {
...
-    let serviceType = 'Unknown service';
-    let icon = 'unknown';
-
-    switch (this.props.type) {
-      case 'Extended color light':
-        serviceType = 'Light';
-        icon = 'light';
-        break;
-    }
+  renderLightService(isConnected) {
```
azasypkin: Don't have better idea right now, how to split or generalize service item rendering without big changes (I'd defer something big for later). What do you think?

azasypkin:
f? @gmarty 

You can't really test it right now, but I'd rather start discussion (you see I have a lot of questions) earlier than later :)

app/index.html
gmarty: The CSP is static, so `img-src *` is good.
You may also want to alter the `media-src` directive if you want to stream video from the box.

app/js/lib/foxbox.js
gmarty: Constants should be moved to `foxbox-settings.js`. See how `REGISTRATION_SERVICE` is implemented for an example.

app/js/lib/foxbox.js
gmarty: We want to reject the promise if the response is not OK, rather than resolving data that the app can't process.

app/js/lib/foxbox.js
gmarty: I don't think we should do that.
The goal of this lib is to provide an abstraction layer over the communication protocol. I think we need to implement methods like `getSnapshot` or `getVideoStream` that accept the service ID as a parameter and that return whatever the view needs.

app/js/controllers/service.js
gmarty: I'd rather have that logic moved to a view (i.e. `service.jsx`), then the

app/js/lib/foxbox.js
azasypkin: Mmm, but ```getSnapshot``` is too specific to one of the tons (hopefully) adapters we're going to have. Shouldn't we just have separate ```controllers``` that extend foxbox? Like IP camera controller, Hue controller, Cast controler and etc. Or what is the idea?

app/js/controllers/service.js
azasypkin: Do you have any example when view itself can render another view to demonstrate what you mean? Sorry I'm quite new React JS?

app/index.html
azasypkin: Okay, good and nope, no video for now in upstream :/

app/css/vendors/ip-camera.css
gmarty: I was thinking vendors == manufacturers. I think we need a `css/services/` folders for these styles.

app/js/controllers/service.js
azasypkin: Ah, okay, will read about these lifecycle methods. So let me try to do something like this.

app/css/vendors/ip-camera.css
gmarty: I  don't usually have opinions on CSS. If you think that's the best way to do, implement it, update the existing styles and document the best practices in `app/css/README.md`.

app/css/vendors/ip-camera.css
gmarty: No you're correct, the header styling needs some love and fixing.
Can you do that for all pages instead?

app/js/controllers/service.js
gmarty: I think something like that might work:
```js
render() {
 switch(this.state.service) {
  case 'ipcamera':
    return (<IpCamera ...props/>);
  default:
    return (<DefaultService ...props/>);
 }
}
```

app/js/lib/foxbox.js
gmarty: I want the lib to be easy to understand by regular developers. The long-term of the lib is to be independent from the app and live in its own repo so that 3rd party dev can reuse it.
If they have to look into the API of every single service they want to use, that defeats the point of having a lib in the first place.

I understand that adding many many specific API is cumbersome and we need a better approach.
How about an object oriented approach? You get an instance of `IpCamera` object from the `foxbox.getServices()` and this object has specific API like `.getSnapshot()` or `.getVideoStream()`. This way we don't have to pass service IDs, just to keep a reference to objects.
What do you think?

app/js/views/camera-service.jsx
gmarty: Yes, I commented about that above and I think it makes sense too.

app/css/vendors/ip-camera.css
azasypkin: Okay, let's try this then

app/css/vendors/ip-camera.css
azasypkin: > update the existing styles

I'll check how much work it is, if not much - will do in this pr, otherwise - I'll handle this in the follow up to not mix things too much.

app/css/vendors/ip-camera.css
azasypkin: > Can you do that for all pages instead?

Looks doable even for this PR, let me try.

app/js/controllers/service.js
azasypkin: Will give it a shot, thanks!

app/js/lib/foxbox.js
azasypkin: Yep, makes sense

app/js/lib/foxbox.js
azasypkin: Yeah, but ```resp.ok()``` is already checked in ```fetchJSON``` and if ```!resp``` is checked here (we throw - it's the same as ```reject```)?

app/js/lib/foxbox.js
azasypkin: Yeah, I see your point, it sounds like something that deserves a separate PR and a bit of discussion - we'll likely have time in MTV to do this :) Foxbox lib should provide some sort of extension point so that anything like ```getSnapshot``` or ```castMyImage``` won't require changes in the lib and at the same time lib will provide useful functionality for these methods (eg. adding auth headers).

So would be it OK for you to live with ```performServiceCommand``` in this PR + I file and handle the issue to make it more generic and nice as we'll discuss in MTV ? :)

app/css/app.css
```diff
@@ -86,81 +54,50 @@ button {
   border-radius: .3rem;
 }
 
-/* Modal */
-.modal {
```
azasypkin: Since ```modal``` component is not used, I've just moved it's styles to a separate file (I think later one we'll just add build step to concatenate some files for example) and didn't do any changes.

app/css/components/footer-menu.css
```diff
@@ -0,0 +1,52 @@
+/* Footer menu */
+.footer-menu {
+  display: flex;
+
+  width: 100%;
+  height: 100%;
+}
+
+.footer-menu__item {
```
azasypkin: Once you're fine with new naming and styles, I'll add note to the readme about the final style.

app/js/views/services/default-service.jsx
```diff
@@ -0,0 +1,86 @@
...
+    this.foxbox = props.foxbox;
+    this.service = props.service;
+
+    this.onServiceStateChanged = this.onServiceStateChanged.bind(this);
+  }
+
+  componentDidMount() {
+    this.populateTags();
+
+    this.foxbox.addEventListener(
```
azasypkin: I believe we have an issue in master, I didn't try to fix it in this PR. The issue is that we don't analyze service id in ```service-state-change``` event handler, so that if service1 changed state, then service2 that has this listener will override its state by the state of service2.

azasypkin:
Aaaaand I think it's ready for review so r? @gmarty 

I've left some explanations inline.

app/index.html
```diff
@@ -2,17 +2,22 @@
...
 <link rel="stylesheet" href="css/FiraSans/font.css">
-<link rel="stylesheet" href="css/app.css">
-<link rel="stylesheet" href="css/user-login.css">
-<link rel="stylesheet" href="css/services.css">
+<link rel="stylesheet" href="css/components/footer-menu.css">
+<link rel="stylesheet" href="css/components/modal.css">
+<link rel="stylesheet" href="css/components/service-list.css">
+<link rel="stylesheet" href="css/components/tag-list.css">
+<link rel="stylesheet" href="css/views/services/default-service.css">
+<link rel="stylesheet" href="css/views/services/ip-camera-service.css">
```
gmarty: It's already in `services/`, so let's just call it `ip-camera.css`.

app/css/app.css
```diff
@@ -1,52 +1,20 @@
...
+
   font: normal 10px/10px Arial;
 }
 
 body {
+  height: 100%;
   margin: 0;
   padding: 0;
+
+  background: #FFF;
```
gmarty: Please lower case for colours, or better in this case, use `white` (easier to read and understand for us poor human being)

app/css/components/footer-menu.css
gmarty: As I said I don't really have opinions on style naming, so I leave this part entirely to you.
I'm happy as long as it's documented (which can wait on a subsequent pull request).

app/index.html
```diff
@@ -22,6 +27,6 @@
 <script defer src="js/initapp.js"></script>
 </head>
 <body>
-<section id="main" class="active"></section>
+  <section class="app-view-container"></section>
```
gmarty: nit: DOM tree is pretty shallow, no need to indent immediate children.

app/js/lib/foxbox.js
gmarty: Yes, you're absolutely right.

app/js/lib/foxbox.js
gmarty: OK, let's leave it for now as a workaround until "`frontend adapters`" land. They'll be part of the presentation I'll give during the workweek, but let's make sure we specify them and come with a plan regarding the implementation.
I do think that the lib should be kept in sync with the box so that we don't need to extend it. I believe the 3rd party dev should completely ignore the underlying REST API and only work with a user friendly, high level API.
Exposing such a low-level method may harm retrocompatibility in the future and block us from making any changes to the REST API implemented in the box.
So in the meantime, please add a big comment saying something like `Do not use! Temporary fix will be deprecated very soon` and show this warning in a `console.warn()` too.

app/js/views/service.jsx
```diff
@@ -1,82 +1,35 @@
...
-        <FooterMenu/>
-      </div>
-    );
+    switch(this.state.service.properties.type) {
+      case 'ipcamera':
+        return (<CameraService service={this.state.service}
+                               foxbox={this.foxbox} />);
+      default:
+        return (<DefaultService service={this.state.service}
+                                foxbox={this.foxbox} />);
```
gmarty: Is there anything from the layout that we can move from the services components to here to make more them more compact?
For example, can we move the header or the footer menu to this component.

app/js/views/services-list-item.jsx
```diff
@@ -149,21 +147,42 @@ export default class ServicesListItem extends React.Component {
...
+
+  render() {
+    switch (this.props.type) {
+      case 'Extended color light':
+        return this.renderLightService();
+      case 'ipcamera':
+        return this.renderGenericService('Camera', 'ip-camera');
+      default:
+        return this.renderGenericService();
+    }
```
gmarty: We probably want to refactor that on the model of `views/services/` and have `views/services/ligh-list-item.jsx` and `views/services/camera-list-item.jsx`.
But this can be done later.

app/js/views/services/default-service.jsx
gmarty: Yes, you're right. `onServiceStateChanged` must check that the ID of the updated service is the same as the current service one. Let's fix it later.

gmarty:
I finished another round of review, but I didn't really look at the CSS changes and I want to test it locally before merging.
I left some nits and comments.

app/js/views/services/camera-service.jsx
```diff
@@ -0,0 +1,90 @@
...
+          <section className="camera-controls__snapshot-tools">
+            <button className="camera-controls__snapshot-btn" type="button"
+                    title="Take a snapshot"
+                    onClick={this.takeSnapshot.bind(this)}>
+            </button>
+            <img ref="previousSnapshot"
+                 className="camera-controls__previous-snapshot" />
+          </section>
+        </div>
+        <FooterMenu className="app-view__footer" />
```
gmarty: Why is the different class needed on some pages only?

app/css/app.css
azasypkin: Regarding case: definitely, lower case is much better!
Regarding aliases: honestly I'm against ```white``` since it's very limited set, so some colors will be in that notation, some will be definitely in a hex + it's much easier to copy-paste to visual editors since they all understand hex notation, but not human-readable aliases.

azasypkin: Btw, I really like https://google.github.io/styleguide/htmlcssguide.xml + BEM naming + properties grouping (https://github.com/necolas/idiomatic-css#declaration-order), not sure what you think.

app/index.html
azasypkin: Okay!

Just a bit of history to tell why I named it so, just as a fun story :) Some people use editors that don't have project explorer and don't indicate full path for the opened file. So they can open files with the same name, but from different directories - like ```services/ip-camera.js``` and ```controllers/ip-camera.js``` :)

app/js/lib/foxbox.js
azasypkin: > OK, let's leave it for now as a workaround until "frontend adapters" land. They'll be part of the presentation I'll give during the workweek, but let's make sure we specify them and come with a plan regarding the implementation.

Wow, eager to hear!

>So in the meantime, please add a big comment saying something like Do not use! Temporary fix will be deprecated very soon and show this warning in a console.warn() too.

Sure.

app/js/views/service.jsx
azasypkin: Yep, as we discussed let's do this in a separate PR, I've left it as close to the master as possible (every view builds UI from header to footer).

app/js/views/services-list-item.jsx
azasypkin: Yep, agree for both statements :)

PR #14, opened 2016-03-16
Make tests clean up after they are executed

JohanLorenzo:


JohanLorenzo:
LocalStorage and IndexedDb are not preserved anymore between tests. r? @gmarty

app/js/lib/foxbox.js
```diff
@@ -89,9 +89,19 @@ export default class Foxbox extends Service {
           this.togglePolling(settings.pollingEnabled);
         });
         this.togglePolling(settings.pollingEnabled);
+      }).then(() => {
+        window.foxbox = this;
```
gmarty: Can you move this to the top of `init`? No need to wrap it in a `then()`.

app/js/lib/foxbox-db.js
```diff
@@ -56,6 +56,17 @@ export default class FoxboxDb {
     }
   }
 
+  drop() {
```
gmarty: Naming is hard, but what about something like `clear`. That's consistent with that's in `foxbox-db.js`.

app/js/lib/foxbox-settings.js
```diff
@@ -32,6 +32,10 @@ export default class FoxboxSettings extends Model {
     });
   }
 
+  clear() {
+    storage.clear();
```
gmarty: We really just want to remove the items set here so that in the future this lib can be embedded in 3rd party projects without having side effects.
It's probably good enough for now, but I'd really add a comment like:
```js
// @todo Remove only the items set here.
```

Lastly, in an ideal world, we want to wrap it in a promise. Local storage is known for blocking the main thread because it's a synchronous API. With a promise we would get the benefit of a consistent API with `foxbox-db.js` too.

app/js/lib/foxbox.js
```diff
@@ -89,9 +89,19 @@ export default class Foxbox extends Service {
...
       });
   }
 
+  /**
+   * Clear all data/settings stored on the browser. Use with caution.
+   */
+  clearAll() {
+    settings.clear();
+    return db.drop();
```
gmarty: When/If `settings.clear()` returns a promise, we can wrap it all in a call to `Promise.all` and return it directly.

app/js/lib/foxbox.js
```diff
@@ -89,9 +89,19 @@ export default class Foxbox extends Service {
...
         this.togglePolling(settings.pollingEnabled);
+      }).then(() => {
+        window.foxbox = this;
       });
   }
 
+  /**
+   * Clear all data/settings stored on the browser. Use with caution.
+   */
+  clearAll() {
```
gmarty: Also `clear()` for the sake of consistency.

JohanLorenzo:
All comments have been addressed: `clear` and promises are used everywhere, windox.foxbox is defined outside of its own then.

app/js/lib/foxbox.js
```diff
@@ -92,6 +93,14 @@ export default class Foxbox extends Service {
       });
   }
 
+  /**
+   * Clear all data/settings stored on the browser. Use with caution.
+   */
+  clear() {
+    let promises = [settings, db].map(el => el.clear());
+    return Promise.all(promises);
```
gmarty: This is easier to read:
```js
const promises = [settings.clear(), db.clear()];
return Promise.all(promises);
```

PR #16, opened 2016-03-17
Schedule next poll only once the previous one is finished + small refactoring to have less iterations through service list.

azasypkin:
r? @gmarty 

Hey @gmarty, yesterday Fabrice asked me to make polling a bit less aggressive for yesterday's demo. I think it may beneficial to land reduced version of my yesterday's patch as well:

* Schedule next polling only once all requests from previous attempt have completed as ```setInterval``` can bump server with requests quite easily in slow network conditions;
* Reduced number of iterations through service collections.

I still see the issue (same on master), when service doesn't appear after it has disappeared, but it's another story.

And maybe 2sec polling is too frequent? Don't know really.

Does PR look sane to you? Have I missed anything?

### DIFF
--- a/app/js/lib/foxbox.js
+++ b/app/js/lib/foxbox.js
@@ -19,7 +19,7 @@ let _remote = false;
 // A reference to the interval to get the online status.
 let _onlineInterval = null;
 
-const privateMembers = {
+const _private = {
   isPollingEnabled: Symbol('isPollingEnabled'),
   nextPollTimeout: Symbol('nextPollTimeout')
 };
@@ -321,14 +321,14 @@ export default class Foxbox extends Service {
    * be started or stopped.
    */
   togglePolling(pollingEnabled) {
-    this[privateMembers.isPollingEnabled] = pollingEnabled;
+    this[_private.isPollingEnabled] = pollingEnabled;
 
     if (pollingEnabled) {
       this.schedulePoll();
     } else {
       // Cancel next poll attempt if it has been scheduled.
-      clearTimeout(this[privateMembers.nextPollTimeout]);
-      this[privateMembers.nextPollTimeout] = null;
+      clearTimeout(this[_private.nextPollTimeout]);
+      this[_private.nextPollTimeout] = null;
     }
   }
 
@@ -339,18 +339,18 @@ export default class Foxbox extends Service {
    */
   schedulePoll() {
     // Return early if polling is not enabled or it has already been scheduled.
-    if (!this[privateMembers.isPollingEnabled] ||
-        this[privateMembers.nextPollTimeout]) {
+    if (!this[_private.isPollingEnabled] ||
+        this[_private.nextPollTimeout]) {
       return;
     }
 
-    this[privateMembers.nextPollTimeout] = setTimeout(() => {
+    this[_private.nextPollTimeout] = setTimeout(() => {
       this.refreshServicesByPolling()
         .catch((e) => {
           console.error('Polling has failed, scheduling one more attempt: ', e);
         })
         .then(() => {
-          this[privateMembers.nextPollTimeout] = null;
+          this[_private.nextPollTimeout] = null;
 
           this.schedulePoll();
         });
--- a/app/js/lib/foxbox.js
+++ b/app/js/lib/foxbox.js
@@ -19,7 +19,7 @@ let _remote = false;
 // A reference to the interval to get the online status.
 let _onlineInterval = null;
 
-const priv = {
+const privateMembers = {
   isPollingEnabled: Symbol('isPollingEnabled'),
   nextPollTimeout: Symbol('nextPollTimeout')
 };
@@ -321,14 +321,14 @@ export default class Foxbox extends Service {
    * be started or stopped.
    */
   togglePolling(pollingEnabled) {
-    this[priv.isPollingEnabled] = pollingEnabled;
+    this[privateMembers.isPollingEnabled] = pollingEnabled;
 
     if (pollingEnabled) {
       this.schedulePoll();
     } else {
       // Cancel next poll attempt if it has been scheduled.
-      clearTimeout(this[priv.nextPollTimeout]);
-      this[priv.nextPollTimeout] = null;
+      clearTimeout(this[privateMembers.nextPollTimeout]);
+      this[privateMembers.nextPollTimeout] = null;
     }
   }
 
@@ -339,18 +339,21 @@ export default class Foxbox extends Service {
    */
   schedulePoll() {
     // Return early if polling is not enabled or it has already been scheduled.
-    if (!this[priv.isPollingEnabled] || this[priv.nextPollTimeout]) {
+    if (!this[privateMembers.isPollingEnabled] ||
+        this[privateMembers.nextPollTimeout]) {
       return;
     }
 
-    this[priv.nextPollTimeout] = setTimeout(() => {
-      this.refreshServicesByPolling().catch((e) => {
-        console.error('Polling has failed, scheduling one more attempt: ', e);
-      }).then(() => {
-        this[priv.nextPollTimeout] = null;
+    this[privateMembers.nextPollTimeout] = setTimeout(() => {
+      this.refreshServicesByPolling()
+        .catch((e) => {
+          console.error('Polling has failed, scheduling one more attempt: ', e);
+        })
+        .then(() => {
+          this[privateMembers.nextPollTimeout] = null;
 
-        this.schedulePoll();
-      });
+          this.schedulePoll();
+        });
     }, settings.pollingInterval);
   }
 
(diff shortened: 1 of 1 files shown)


app/js/lib/foxbox.js
```diff
@@ -10,6 +10,11 @@ import FoxboxDb from './foxbox-db';
 const settings = new FoxboxSettings();
 const db = new FoxboxDb();
 
+const priv = {
```
gmarty: Does `priv` stand for `private`? In this case, let's call it `private`.

app/js/lib/foxbox.js
```diff
@@ -192,21 +197,44 @@ export default class Foxbox extends Service {
...
+   * previous one is completed or failed.
+   */
+  schedulePoll() {
+    // Return early if polling is not enabled or it has already been scheduled.
+    if (!this[priv.isPollingEnabled] || this[priv.nextPollTimeout]) {
+      return;
+    }
+
+    this[priv.nextPollTimeout] = setTimeout(() => {
+      this.refreshServicesByPolling().catch((e) => {
```
gmarty: Please add a line break before `.catch(...`.

app/js/lib/foxbox.js
```diff
@@ -192,21 +197,44 @@ export default class Foxbox extends Service {
...
+        console.error('Polling has failed, scheduling one more attempt: ', e);
+      }).then(() => {
```
gmarty: Please add a line break before `.then(...`.
(See the rest of the code for coding styles)

app/js/lib/foxbox.js
```diff
@@ -218,78 +246,57 @@ export default class Foxbox extends Service {
...
-                const storedService = storedServices.find(s => s.id === fetchedService.id);
-
-                if (!storedService) {
-                  this._dispatchEvent('service-state-change', fetchedService);
-
-                  // Populate the db with the latest service.
-                  db.setService(fetchedService);
+    let fetchedServicesPromise = fetchJSON(`${this.origin}/services/list`).then(
+      (services) => {
+        // TODO: We should ask for state only for services that actually support
```
gmarty: Please use the `// @todo We should...` form.

app/js/lib/foxbox.js
```diff
@@ -218,78 +246,57 @@ export default class Foxbox extends Service {
...
-                fetchedService.state = states[id];
-
-                const storedService = storedServices.find(s => s.id === fetchedService.id);
-
-                if (!storedService) {
-                  this._dispatchEvent('service-state-change', fetchedService);
-
-                  // Populate the db with the latest service.
-                  db.setService(fetchedService);
+    let fetchedServicesPromise = fetchJSON(`${this.origin}/services/list`).then(
```
gmarty: Same here.

app/js/lib/foxbox.js
```diff
@@ -218,78 +246,57 @@ export default class Foxbox extends Service {
...
+              (state) => service.state = state
+            );
+          })
+        ).then(() => services);
+      }
+    );
+
+    return Promise.all([
+      this.getServices(),
+      fetchedServicesPromise
```
gmarty: If I understand correctly, `fetchedServicesPromise` contains value and not a promise. Is that right?
If so, we want to run both promises synchronously in the `Promise.all()`.

app/js/lib/foxbox.js
```diff
@@ -218,78 +246,57 @@ export default class Foxbox extends Service {
...
+    ]).then(([storedServices, fetchedServices]) => {
+      let hasNewServices = fetchedServices.reduce(
+        (hasNewServices, fetchedService) => {
+          const storedService = storedServices.find(
```
gmarty: Why not directly do:
```js
const isExistingService = !!storedServices.find(...
```

app/js/lib/foxbox.js
```diff
@@ -218,78 +246,57 @@ export default class Foxbox extends Service {
...
+          fetchedService = isExistingService ?
+            Object.assign(storedService, fetchedService) : fetchedService;
 
-                if (!isSimilar(fetchedService.state, storedService.state)) {
-                  fetchedService = Object.assign(storedService, fetchedService);
+          this._dispatchEvent('service-state-change', fetchedService);
 
-                  this._dispatchEvent('service-state-change', fetchedService);
+          // Populate the db with the latest service.
+          db.setService(fetchedService);
```
gmarty: You need to add the line with `Object.assign`. Currently the tags associated to a service are stored as a property in the service object in the database. We want to keep these and any other extended values on the updated object and store it in the database.

app/js/lib/foxbox.js
```diff
@@ -218,78 +246,57 @@ export default class Foxbox extends Service {
...
+          }
 
-                  return;
-                }
+          fetchedService = isExistingService ?
+            Object.assign(storedService, fetchedService) : fetchedService;
 
-                if (!isSimilar(fetchedService.state, storedService.state)) {
-                  fetchedService = Object.assign(storedService, fetchedService);
+          this._dispatchEvent('service-state-change', fetchedService);
```
gmarty: I think we should only emit a `service-state-change` event when an existing service state changed, not when a new service is connected to the box.

app/js/lib/foxbox.js
```diff
@@ -218,78 +246,57 @@ export default class Foxbox extends Service {
...
+          return hasNewServices || !isExistingService;
+        },
+        false /* hasNewServices */
+      );
 
-              if (hasChanged) {
-                // The state of the services changes.
-                this._dispatchEvent('service-change', fetchedServices);
-              }
+      if (hasNewServices || fetchedServices.length !== storedServices.length) {
```
gmarty: Is this going to work if between 2 successive polling a new service is connected and a service disconnected?
Are disconnected services properly taking into account?

app/js/lib/foxbox.js
azasypkin: Yep, it's ```private`` just used convention from SMS team, but I'm fine with ```private``` too - will rename it :)

app/js/lib/foxbox.js
azasypkin: Mmm, but isn't it the same we have in master [1]?

If so, I'd rather keep master behavior here for now and will be happy to change it in a separate PR

[1] https://github.com/fxbox/app/blob/master/app/js/lib/foxbox.js#L275-L276

app/js/lib/foxbox.js
azasypkin: Ugh, need to get used to this convention :) Will change

app/js/lib/foxbox.js
azasypkin: Ok!

app/js/lib/foxbox.js
azasypkin: I think so, let me think aloud:
* If new service connected we'll have ```hasNewServices === true``` since we iterate through all _latest_ services and if some are new - we'll have this flag set, independently from any disconnected services;
* If old service disconnected, we'll have different length;
* If new service connected and old is disconnected we'll have ```hasNewServices === true``` anyway.

Am I missing something?

app/js/lib/foxbox.js
azasypkin: Ha, I've remembered why we called it ```priv``` - ```private``` is reseved keyword in strict mode [1] :)

Do you want me to rename it to something like ```privateMembers``` then or just leave ```priv```?

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Paving_the_way_for_future_ECMAScript_versions.

azasypkin:
@gmarty added second commit with changes and replied to two remaining questions. How does it look?

Thanks!

app/js/lib/foxbox.js
```diff
@@ -10,6 +10,11 @@ import FoxboxDb from './foxbox-db';
 const settings = new FoxboxSettings();
 const db = new FoxboxDb();
 
+const privateMembers = {
```
gmarty: nit: `privateMembers` is a bit too long, why not just `private`? (sorry)

app/js/lib/foxbox.js
```diff
@@ -201,18 +206,44 @@ export default class Foxbox extends Service {
...
 
-      // We immediately sync the data with the box.
-      this.refreshServicesByPolling();
+    if (pollingEnabled) {
+      this.schedulePoll();
     } else {
-      clearInterval(this.pollingInterval);
+      // Cancel next poll attempt if it has been scheduled.
+      clearTimeout(this[privateMembers.nextPollTimeout]);
+      this[privateMembers.nextPollTimeout] = null;
```
gmarty: Do you really need to set it to `null` after calling `clearTimeout`?

app/js/lib/foxbox.js
azasypkin: Mmm, I've already asnwered why ```private``` is not possible :) [1]

[1] https://github.com/fxbox/app/pull/16#discussion_r56576304

azasypkin: Mmm, ok I can make it ```_private``` :)

app/js/lib/foxbox.js
azasypkin: Otherwise next call to ```schedulePoll``` will do early return

azasypkin: That can happen when we turned off polling and later want to enable it.

gmarty:
That looks really good! I just left a nit and a general question.
I think you need to rebase because the master has changed a lot lately.

azasypkin:
> That looks really good! I just left a nit and a general question.
I think you need to rebase because the master has changed a lot lately.

Yep, rebased, renamed to ```_private``` and answered question :)

PR #17, opened 2016-03-17
testing logout

isabelrios:
Just a very easy test to start with

### DIFF
--- a/tests/common/logout_test.js
+++ b/tests/common/logout_test.js
@@ -1,0 +1,25 @@
+'use strict';
+
+var App = require('../lib/app');
+
+describe('Logout', function() {
+  var app;
+  var loginView;
+  var servicesView;
+  this.timeout(30000);
+
+  before(() => {
+    app = new App();
+    return app.init()
+      .then(defaultView => { loginView = defaultView; });
+  });
+
+  after(() => {
+    return app.stop();
+  });
+
+  it('should logout', () => {
+    return loginView.loginSuccess(12345678)
+      .then(servicesView => servicesView.logoutSuccess());
+  });
+});
--- a/tests/lib/views/login/view.js
+++ b/tests/lib/views/login/view.js
@@ -5,6 +5,8 @@ var LoginAccessors = require('./accessors');
 function LoginView() {
   [].push.call(arguments, LoginAccessors);
   View.apply(this, arguments);
+  
+  this.accessors.startLoginButton;
 }
 
 LoginView.prototype = Object.assign({
@@ -14,7 +16,8 @@ LoginView.prototype = Object.assign({
       this.accessors.passwordField.sendKeys(password);
       this.accessors.submitButton.click();
     }).then(() => {
-      return new require('../services/view')(this.driver);
+      var ServicesView = require('../services/view');
+      return new ServicesView(this.driver);
     });
   }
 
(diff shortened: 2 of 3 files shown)


isabelrios:
I created the test as a separete one but could change that. JohanLorenzo r?

tests/common/logout_test.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+var App = require('../lib/app');
+var assert = require('assert');
```
JohanLorenzo: Nit: unused import

tests/common/logout_test.js
```diff
@@ -0,0 +1,30 @@
...
+    return app.init()
+      .then(defaultView => { loginView = defaultView; });
+  });
+
+  after(() => {
+    return app.stop();
+  });
+
+  it('should logout', () => {
+    return loginView.loginSuccess(12345678).then(function(servicesView) {
```
JohanLorenzo: Nit: `.then()` shouldn't be nested.

tests/lib/views/login/view.js
```diff
@@ -14,8 +14,13 @@ LoginView.prototype = Object.assign({
       this.accessors.passwordField.sendKeys(password);
       this.accessors.submitButton.click();
     }).then(() => {
-      return new require('../services/view')(this.driver);
+      var ServicesView = require('../services/view');
+      return new ServicesView(this.driver);
     });
+  },
+
+  isLoginButton: function() {
```
JohanLorenzo: Views shouldn't expose HtmlElements. How about moving `this.accessors.startLoginButton` in the constructor of LoginView? Then when when `LoginView` [is created here](https://github.com/fxbox/app/pull/17/files#diff-da27a7816cd34a151cc34d117ce6cf7eR17), we're sure the login button is visible.

tests/common/logout_test.js
```diff
@@ -0,0 +1,30 @@
...
+      return  servicesView.logoutSuccess().then(function(loginView) {
+        return loginView.isLoginButton();
```
JohanLorenzo: As a consequence of [this comment](https://github.com/fxbox/app/pull/17/files#r56498382), this line shoudn't be needed anymore.

JohanLorenzo:
It looks nice! :smiley:

r+ modulo [this change](https://github.com/fxbox/app/pull/17/files#r56498382).

tests/common/logout_test.js
isabelrios: So it cannot be nested inside the arrow funtion?
Would this be valid, changing how to start the test?
it('should logout', function() {

Otherwise I do not know how not to nest the .then...:S

JohanLorenzo: It can be nested, but it's not considered as a good practice. Promises were thought as an improvement of callbacks, and nesting was [one of the pain point](http://callbackhell.com/).

The way to prevent nesting is the following:
```js
return loginView.loginSuccess(12345678)
  .then(servicesView => {
    return servicesView.logoutSuccess(); // promise is returned there and chained to the top level.
  }).then(() => {
    // do something after it's logged out
  })
```

Thanks to arrow functions, this can be shortened to:
```js
return loginView.loginSuccess(12345678)
  .then(servicesView => servicesView.logoutSuccess())
  .then(() => {
    // do something after it's logged out
  })
```

isabelrios: Sorry thought I could not use the first .then inside the first arrow function. The first example is the same like it was but there is one more then. I understand know what you mean, in case there were more how it is best to handle them.
Thanks!

PR #23, opened 2016-03-21
[WIP] First approach to feed foxapp

isabelrios:
The idea is to simulate the app having services for the automated tests


azasypkin:
Hey @isabelrios is this PR still in progress/needed? Do you need any help/review here?

Nothing urgent, it's just my spring PR/issue cleanup :)

isabelrios:
Hi @azasypkin, we can close it. I should have close it before, since there was not activity on this lately, have to come back to this but with all recent changes and the way we are going to approach this, maybe it would make sense to create a new one.Thanks!

PR #25, opened 2016-03-23
Web Push (tested on Firefox)

arcturus:
So far just print to the console the endpoint and the key, we will need to send it to the
server once it's implemented there.

This is just a WIP

@gmarty @aosmond thoughts?


gmarty:
Is it possible to decouple the code from the app and move the logic in the `foxbox.js` lib?
Ideally, the push listener would emit an event and the app would be responsible to display the notification.

Would that even work?

arcturus:
I see some code going to the library, but actually that code is not in this PR :)

This PR lacks of the need of saving the endpoint + keys, something that we need to send to the box, and that's makes perfect sense to be in the foxbox.js lib.

The rest of changes I'm not sure if they should be on the library itself:
- hookSW.js: is a piece of js that will be running in the sw itself. In charge of listen to the notifications and **showing** them. Remember that sometimes your app wont be running and you'll display the notification from the SW. One thing we can do is, if the app is running, send the notification back via post message and perhaps using the lib to display notifications.
- notification.js: could be in any other file, as this is the part where we tell the sw to subscribe for notifications. This happens in content side (not sw side). But coupling to the library don't know if makes sense.
- the rest of the changes are the gulp task and manifest needed to run sw automation.

app/js/lib/foxbox.js
```diff
@@ -537,4 +537,31 @@ export default class Foxbox extends Service {
   setTag() {
     return db.setTag.apply(db, arguments);
   }
+
+  subscribeForNotifications(resubscribe = false) {
```
gmarty: nit: Maybe `subscribeToNotifications` sounds better.

app/js/lib/foxbox.js
```diff
@@ -537,4 +537,31 @@ export default class Foxbox extends Service {
...
+    if (!navigator.serviceWorker) {
+      return Promise.reject('No service worker supported');
+    }
+
+    if (settings.pushEndpoint && settings.pushPubKey && !resubscribe) {
+      return Promise.resolve();
+    }
+
+    return navigator.serviceWorker.ready.then((reg) => {
+      reg.pushManager.subscribe({userVisibleOnly: true}).then((subscription) => {
```
gmarty: nit: can you insert a line break before `.then(...`? See the rest of the file for coding style.

app/js/lib/foxbox.js
```diff
@@ -537,4 +537,31 @@ export default class Foxbox extends Service {
...
+        const endpoint = subscription.endpoint;
+        const key = subscription.getKey ? subscription.getKey('p256dh') : '';
+        settings.pushEndpoint = endpoint;
+        settings.pushPubKey = btoa(String.fromCharCode.apply(null, new Uint8Array(key)));
+        console.info('Endpoint: ', db.pushEndpoint);
+        console.info('PubKey: ', db.pushPubKey);
+      }).catch((error) => {
```
gmarty: nit: idem.

arcturus:
@gmarty r?

app/hookSW.js
```diff
@@ -0,0 +1,40 @@
...
+});
+
+self.addEventListener('pushsubscriptionchange', function(evt) {
+  console.log('Got subscription change event ', evt);
+});
+
+function showNotification(obj, evt) {
+  var title = 'Link: Notification';
+  var body = obj.body;
+  var icon = 'img/icons/128.png';
```
gmarty: Can we try to use the SVG icon located in `img/icon.svg` instead?

app/hookSW.js
```diff
@@ -0,0 +1,40 @@
+/**
```
gmarty: **Note:** In the future, we will want to write this file in es2015 and add it to the transpilation workflow like the files in the `js/` folder. It will be better if we want to make a es5 build one day. This file needs to be linted as well.
Ping @azasypkin.

app/hookSW.js
```diff
@@ -0,0 +1,40 @@
...
+  var body = obj.body;
+  var icon = 'img/icons/128.png';
+  var tag = 'link-push';
+
+  evt.waitUntil(self.registration.showNotification(title, {
+    body: body,
+    icon: icon,
+    tag: tag
+  }));
+}
```
gmarty: nit: Add a line break at the end of files.

app/js/lib/foxbox/foxbox.js
```diff
@@ -389,6 +389,71 @@ export default class Foxbox extends Service {
     return this[p.db].setTag.apply(this[p.db], arguments);
   }
 
+  subscribeToNotifications(resubscribe = false) {
```
gmarty: Can you annotate this method, specifically to describe what is the `resubscribe` param for?

app/js/lib/foxbox/foxbox.js
```diff
@@ -389,6 +389,71 @@ export default class Foxbox extends Service {
...
+    const boxHost = this[p.net].origin;
+    if (!navigator.serviceWorker) {
+      return Promise.reject('No service worker supported');
+    }
+
+    if (settings.pushEndpoint && settings.pushPubKey && !resubscribe) {
+      return Promise.resolve();
+    }
+
+    return navigator.serviceWorker.ready.then((reg) => {
```
gmarty: nit: Can you add a line break and indent before `.then(`?

app/js/lib/foxbox/foxbox.js
```diff
@@ -389,6 +389,71 @@ export default class Foxbox extends Service {
...
+      reg.pushManager.subscribe({userVisibleOnly: true})
+      .then((subscription) => {
+        const endpoint = subscription.endpoint;
+        const key = subscription.getKey ? subscription.getKey('p256dh') : '';
+        settings.pushEndpoint = endpoint;
+        settings.pushPubKey = btoa(String.fromCharCode.apply(null,
+          new Uint8Array(key)));
+
+        // Send push information to the server
+        // XXX: We will need some library to write taxonomy messages
```
gmarty: nit: Please make this a `// @todo We will need...` comment.

app/js/lib/foxbox/foxbox.js
```diff
@@ -389,6 +389,71 @@ export default class Foxbox extends Service {
...
+              Json: {
+                subscriptions: [{
+                  public_key: settings.pushPubKey,
+                  push_uri: settings.pushEndpoint
+                }]
+              }
+            }
+          ]];
+        return this[p.net].fetchJSON(
+          `${boxHost}/api/v1/channels/set`,
```
gmarty: nit: We use `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/` everywhere else in this file. If it's too big, at least make sure to use the api version from the settings rather than hardcoding it.

app/js/lib/foxbox/foxbox.js
```diff
@@ -389,6 +389,71 @@ export default class Foxbox extends Service {
...
+              }
+
+            ]];
+          return this[p.net].fetchJSON(
+            `${boxHost}/api/v1/channels/set`,
+           'PUT', pushResourcesMsg);
+        });
+      })
+      .catch((error) => {
+        if (Notification.permission === 'denied') {
```
gmarty: **Question:** Are we still getting the push even if the notification permission is denied?

app/js/lib/foxbox/foxbox.js
```diff
@@ -389,6 +389,71 @@ export default class Foxbox extends Service {
...
+        // Send push information to the server
+        // XXX: We will need some library to write taxonomy messages
+        const pushConfigurationMsg = [[
+            [{
+              id: 'setter:subscribe.webpush@link.mozilla.org'
+            }], {
+              Json: {
+                subscriptions: [{
+                  public_key: settings.pushPubKey,
+                  push_uri: settings.pushEndpoint
```
gmarty: Do we really need to store `pushPubKey` and `pushEndpoint` in the settings? It looks like they are always set before being used in the push configuration.
If that's the case, maybe we just need something like `settings.pushConfigured` that defaults to `false`.

gulpfile.js
```diff
@@ -42,7 +43,8 @@ gulp.task('lint', function() {
   return gulp.src([
       'app/**/*.{js,jsx}',
       'tests/**/*.js',
-      '!app/components/**'
+      '!app/components/**',
+      '!app/hookSW.js'
```
gmarty: See note above, we want to lint this file in the future.

gulpfile.js
```diff
@@ -191,8 +193,11 @@ gulp.task('clobber-tests', function() {
  * assets.
  */
 gulp.task('offline', ['build'], function() {
-  gulp.src(['**/*'], { cwd: DIST_APP_ROOT })
-    .pipe(gsww())
+  gulp.src(['**/*'], {cwd: DIST_APP_ROOT})
+    .pipe(gsww({
+      version: package.version,
```
gmarty: **Question:** What is the package version used for? When do we need to bump it?

gmarty:
The patch looks good. I left questions and suggested changes as well as a few nits.
r+
We agreed to test on Chrome and Fennec later.

app/js/lib/foxbox/foxbox.js
arcturus: Nope

gulpfile.js
arcturus: We use it to add it to the serviceworker as a variable.

With that, with each new app version our sw code is different, making the browser to update the SW code, caches, etc.

app/hookSW.js
arcturus: Yes we can, I just checked and it works

app/js/lib/foxbox/foxbox.js
arcturus: Definitely

app/js/lib/foxbox/foxbox.js
arcturus: Sounds good!

PR #26, opened 2016-03-23
Move log out to the footer menu for now

gmarty:


app/js/views/navigation-menu.jsx
```diff
@@ -1,19 +1,29 @@
...
+    this.foxbox = props.foxbox;
+  }
+
   shouldComponentUpdate() {
     // We never need to update this component as it is being recreated each time
     // the route changes.
     return false;
   }
 
+  handleOnClick(evt) {
```
azasypkin: __nit:__ ```evt``` is not used.

app/js/views/navigation-menu.jsx
```diff
@@ -1,19 +1,29 @@
...
+    this.foxbox.logout();
+
+    // Once logged out, we redirect to the login page.
+    location.hash = '#users/login';
```
azasypkin: __nit:__ we can likely just set this hash directly on the link itself, in theory event handler is called first and only then default action is performed.

azasypkin:
r+ with nits addressed and green Travis

gmarty:
Nits addressed and tests fixed.

PR #27, opened 2016-03-24
Change the discovery to use a fingerprint DNS resolvable name rather than an IP address

samgiles:
This needs to land around the same time as https://github.com/fxbox/registration_server/pull/18

### DIFF
--- a/app/js/lib/foxbox/foxbox.js
+++ b/app/js/lib/foxbox/foxbox.js
@@ -97,8 +97,12 @@ export default class Foxbox extends Service {
     return Promise.all(promises);
   }
 
-  get localHostname() {
-    return this[p.settings].localHostname;
+  get localOrigin() {
+    return this[p.settings].localOrigin;
+  }
+
+  get clientId() {
+    return this[p.settings].clientId;
   }
 
   get boxes() {
@@ -106,8 +110,7 @@ export default class Foxbox extends Service {
   }
 
   /**
-   * Get the IP address of the box on the local network using the registration
-   * server.
+   * Get the URL of the box using the registration server.
    * If it fails, we fallback to the previously set hostname.
    * It there isn't, it falls back to localhost.
    *
@@ -134,7 +137,17 @@ export default class Foxbox extends Service {
           this[p.boxes] = Object.freeze(
             boxes
               .filter(box => box.timestamp - now >= 0)
-              .map(box => Object.freeze(box))
+              .map(box => {
+                  // NOTE(sgiles): There is consideration to allow
+                  // only "local_origin" and "tunnel_origin", removing the need
+                  // to parse message - this merges the relevant message fields
+                  // into the main object
+                  const { local_origin, tunnel_origin } =
+                    JSON.parse(box.message);
+                  box.local_origin  = local_origin;
+                  box.tunnel_origin = tunnel_origin;
+                  return Object.freeze(box);
+              })
           );
 
           if (!this[p.settings].configured) {
@@ -172,13 +185,14 @@ export default class Foxbox extends Service {
 
     const box = this[p.boxes][index];
 
-    this[p.settings].localHostname = box.local_ip;
-    if (box.tunnel_url) {
-      this[p.settings].tunnelHostname = box.tunnel_url;
+    this[p.settings].localOrigin = box.local_origin;
+    if (box.tunnel_origin) {
+      this[p.settings].tunnelOrigin = box.tunnel_origin;
     } else {
-      this[p.settings].tunnelHostname = '';
+      this[p.settings].tunnelOrigin = '';
     }
 
+    this[p.settings].clientId = box.client;
     this[p.settings].configured = true;
   }
 
--- a/app/js/lib/foxbox/network.js
+++ b/app/js/lib/foxbox/network.js
@@ -70,15 +70,13 @@ export default class Network {
   get localOrigin() {
     const settings = this[p.settings];
 
-    return `${settings.localScheme}://${settings.localHostname}:` +
-      `${settings.localPort}`;
+    return settings.localOrigin;
   }
 
   get tunnelOrigin() {
     const settings = this[p.settings];
 
-    return `${settings.tunnelScheme}://${settings.tunnelHostname}:` +
-      `${settings.tunnelPort}`;
+    return settings.tunnelOrigin;
   }
 
   get connected() {
@@ -186,7 +184,7 @@ export default class Network {
       });
 
     // @todo Find a better way to detect if a tunnel connection is active.
-    if (this[p.settings].tunnelHostname !== '') {
+    if (this[p.settings].tunnelOrigin !== '') {
       this[p.ping](`${this.tunnelOrigin}/ping`)
         .then(() => {
           this[p.remote] = true;
--- a/app/js/lib/foxbox/settings.js
+++ b/app/js/lib/foxbox/settings.js
@@ -5,14 +5,12 @@ import { Model } from 'components/mvc';
 // Prefix all entries to avoid collisions.
 const PREFIX = 'foxbox-';
 
-const DEFAULT_SCHEME = 'http';
-const DEFAULT_HOSTNAME = 'localhost';
-const DEFAULT_PORT = 3000;
+const DEFAULT_LOCAL_ORIGIN = 'http://localhost:3000';
 const DEFAULT_POLLING_ENABLED = true;
 const POLLING_INTERVAL = 2000;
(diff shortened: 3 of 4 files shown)


app/js/lib/foxbox/foxbox.js
```diff
@@ -97,16 +97,16 @@ export default class Foxbox extends Service {
...
   }
 
   get boxes() {
     return this[p.boxes];
   }
 
   /**
-   * Get the IP address of the box on the local network using the registration
+   * Get the URL of the box using the registration
    * server.
```
azasypkin: __nit:__ please move ```server.``` to the line above as there is enough space for it now :)

app/js/lib/foxbox/foxbox.js
```diff
@@ -134,6 +134,16 @@ export default class Foxbox extends Service {
           this[p.boxes] = Object.freeze(
             boxes
               .filter(box => box.timestamp - now >= 0)
+              .map(box => {
+
```
azasypkin: __nit:__ redundant empty line

app/js/lib/foxbox/foxbox.js
```diff
@@ -134,6 +134,16 @@ export default class Foxbox extends Service {
...
+                  // NOTE(sgiles): There is consideration to allow
+                  // only "local_url" and "tunnel_url", removing the need
+                  // to parse message
+                  let message = JSON.parse(box.message);
```
azasypkin: __nit:__ ```const {local_url, tunnel_url} = JSON.parse(box.message);```

app/js/lib/foxbox/foxbox.js
```diff
@@ -134,6 +134,16 @@ export default class Foxbox extends Service {
...
+                  box.local_url  = message.local_url;
+                  box.tunnel_url = message.tunnel_url;
+                  return box;
```
azasypkin: __optional nit:__ I think it's OK to freeze ```box``` here to avoid one more ```map``` iteration.

app/js/lib/foxbox/network.js
```diff
@@ -56,29 +56,27 @@ export default class Network {
...
+      return this.tunnelUrl;
     }
 
     console.error('The box is out of reach.');
-    return this.localOrigin;
+    return this.localUrl;
   }
 
-  get localOrigin() {
+  get localUrl() {
```
azasypkin: __question:__ Why do we rename ```origin``` to ```URL```? I thought ```local_url``` and ```tunnel_url``` are always just ```origins``` [1] or they can contain any ```path``` part as well? It just looks a bit confusing when we create URLs with something like ```${box.local_url}\path1\path2``` - ```box.local_url``` looks more like ```box.local_origin``` or at least ```box.local_baseURL```

[1] https://tools.ietf.org/html/rfc6454#section-3.2

app/js/lib/foxbox/settings.js
```diff
@@ -83,64 +76,24 @@ export default class Settings extends Model {
...
+    return this._localUrl;
   }
 
-  set tunnelHostname(hostname) {
-    hostname = String(hostname) || DEFAULT_HOSTNAME;
-    this._tunnelHostname = hostname.replace(/\/$/, ''); // Trailing slash.
-    storage.setItem(`${PREFIX}tunnelHostname`, this._tunnelHostname);
+  set localUrl(url) {
+    url = String(url) || DEFAULT_URL;
+    this._localUrl = url.replace(/\/$/, ''); // Trailing slash.
```
azasypkin: If we agree that it's ```origin``` and not ```url``` we can do just ```this._localOrigin = url ? (new URL(url)).origin : DEFAULT_ORIGIN;``` that will guard us from non-valid URL's (at least at development stage) and get rid of trailing slashes :)

app/js/views/user-login.jsx
```diff
@@ -31,13 +31,13 @@ export default class UserLogin extends React.Component {
     if (this.state.boxes.length > 1) {
       let value = this.state.value !== null ? 0 : this.state.value;
       const options = this.state.boxes.map((box, index) => {
-        if (box.local_ip === this.foxbox.localHostname) {
+        if (box.client === this.foxbox.client) {
```
azasypkin: __possible issue:__ what is the ```client``` field? I don't see it anywhere else (e.g. in ```this.foxbox```)?

app/js/views/user-login.jsx
```diff
@@ -31,13 +31,13 @@ export default class UserLogin extends React.Component {
...
           value = index;
         }
-        const label = box.local_ip +
-          (box.tunnel_url ? ` / ${box.tunnel_url}` : '');
+
+        const label = box.client;
```
azasypkin: __nit:__ I'd get rid of ```label``` variable and use ```box.client``` directly in JSX

azasypkin:
Overall changes look good to me, I've just left few nits, one question (I'm not sure if we should rename ```origin``` to ```URL```) and mentioned one possible issue (unknown and suspicious ```this.foxbox.client``` field).

Thanks!

app/js/views/user-login.jsx
samgiles: This is a field that gets returned from the registration server.  It's the fingerprint of the self-signed certificate the box generates.

azasypkin: Yeah, but could you point me to the place where ```this.foxbox.client``` is initialized?

samgiles:
I agree with the `origin`/`url` change, and I've changed the name in https://github.com/fxbox/foxbox/pull/281 to reflect that it should be `origin`. :) thanks, I'll update.

app/js/views/user-login.jsx
samgiles: :D, I see what you meant now.  I've added `clientId` to `foxbox`.

samgiles:
Hey @azasypkin, thanks for the review, I've addressed your nits. Reverted back to the `origin` name, and added clientId to the `foxbox` object.  How does it look now?

samgiles:
Rebased and squashed

app/js/lib/foxbox/network.js
```diff
@@ -56,7 +56,7 @@ export default class Network {
     return Promise.resolve();
   }
 
-  get origin() {
+  get url() {
```
gmarty: Why renaming `origin` by `url`? AFAIK a URL is something like `example.com`, while an origin is `https://example.com:443`.

samgiles:
Just realised I may have merged something that was removed from master back in.

app/js/lib/foxbox/network.js
samgiles: :grimacing: I missed these - it should be `origin`.

app/js/lib/foxbox/settings.js
```diff
@@ -5,14 +5,12 @@ import { Model } from 'components/mvc';
...
-const DEFAULT_SCHEME = 'http';
-const DEFAULT_HOSTNAME = 'localhost';
-const DEFAULT_PORT = 3000;
+const DEFAULT_LOCAL_ORIGIN = 'http://localhost:3000';
 const DEFAULT_POLLING_ENABLED = true;
 const POLLING_INTERVAL = 2000;
 const ONLINE_CHECKING_INTERVAL = 5000;
 const ONLINE_CHECKING_LONG_INTERVAL = 1000 * 60 * 5;
-const REGISTRATION_SERVICE = 'http://knilxof.org:4242/ping';
+const REGISTRATION_SERVICE = 'http://localhost:4242/ping';
```
gmarty: Are we not using the registration server anymore?

app/js/lib/foxbox/foxbox.js
```diff
@@ -134,7 +137,17 @@ export default class Foxbox extends Service {
           this[p.boxes] = Object.freeze(
             boxes
               .filter(box => box.timestamp - now >= 0)
-              .map(box => Object.freeze(box))
+              .map(box => {
+
```
gmarty: nit: line break not needed.

app/js/lib/foxbox/foxbox.js
```diff
@@ -134,7 +137,17 @@ export default class Foxbox extends Service {
...
+                  // NOTE(sgiles): There is consideration to allow
+                  // only "local_origin" and "tunnel_origin", removing the need
+                  // to parse message - this merges the relevant message fields
+                  // into the main object
+                  const {local_origin, tunnel_origin} = JSON.parse(box.message);
```
gmarty: nit: we usually add 1 space within braces: `{ local_origin, tunnel_origin }`.

gmarty:
I left a couple of nits but the rest looks good to me.
r+!

PR #28, opened 2016-03-24
Add the basic recipes section

gmarty:


app/js/controllers/themes.js
```diff
@@ -0,0 +1,29 @@
+/* global React, ReactDOM */
+
+import { Controller } from 'components/fxos-mvc/dist/mvc';
+
+import Themes from 'js/views/themes';
+import ThemesNew from 'js/views/themes-new';
+
+export default class ThemesController extends Controller {
+  main(action = '') {
```
azasypkin: __question:__ mm can't we just remove this default value and use ```default``` switch branch instead of weird ```case ''```?

app/js/views/themes.jsx
```diff
@@ -0,0 +1,94 @@
...
+      }
+
+      return (
+        <li key={id}
+            className={itemClassName}>
+          <input className="themes-list__toggle" type="checkbox"
+                 checked={theme.enabled}
+                 onChange={this.handleOnChange.bind(this, id)}/>
+          <span className="themes-list__name">{theme.name}</span>
+          <img className="themes-list__remove" src="css/icons/clear.svg"
```
azasypkin: __nit:__ can we use ```button``` here? it looks more natural here and better from the accessibility stand point.

app/js/views/themes-new.jsx
```diff
@@ -0,0 +1,221 @@
+/* global React */
+
+import UserLogoutButton from 'js/views/user-logout-button';
```
azasypkin: __nit:__ unused import.

app/js/views/themes-new.jsx
```diff
@@ -0,0 +1,221 @@
...
+
+    this.foxbox = props.foxbox;
+
+    this.updateService = this.updateService.bind(this);
+  }
+
+  componentDidMount() {
+    this.foxbox.getServices()
+      .then(services => {
+        console.log(services);
```
azasypkin: __nit:__ likely forgotten ```console.log``` :)

app/js/views/themes-new.jsx
```diff
@@ -0,0 +1,221 @@
...
+      .catch(console.error.bind(console));
+
+    this.foxbox.addEventListener('service-change', this.updateService);
+  }
+
+  componentWillUnmount() {
+    this.foxbox.removeEventListener('service-change', this.updateService);
+  }
+
+  updateService(services = []) {
```
azasypkin: __nit:__ maybe ```updateServices``` or ```updateServiceList``` since we're updating service list here?

app/js/views/themes-new.jsx
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  handleActionButton() {
+    if (this.state.actionProperty === null) {
+      return;
+    }
+  }
+
+  render() {
+    let className = 'new-theme__select';
```
azasypkin: What about moving parts of this method to a separate methods so that it will be easier to read? Like ```renderServiceSelector```, ```renderPropertySelector``` etc.?

app/js/views/themes-new.jsx
```diff
@@ -0,0 +1,221 @@
...
+    let serviceSelector = (
+      <select value={this.state.service}
+              onChange={this.handleServiceSelection.bind(this)}
+              className={className}>
+        <option value="0">Select a device</option>
+        {optionNodes}
+      </select>
+    );
+
+    className = 'new-theme__select';
```
azasypkin: Not now, but we'll likely need to migrate to https://github.com/JedWatson/classnames someday since we started to manipulate with classes a lot.

app/js/views/themes-new.jsx
```diff
@@ -0,0 +1,221 @@
...
+    if (this.state.actionProperty === null) {
+      actionButtonClassName += ' app-view__action--disabled';
+    }
+
+    return (
+      <div className="app-view">
+        <header className="app-view__header">
+          <a href="#themes" className="app-view__action">Cancel</a>
+          <h1>New Recipe</h1>
+          <a className={actionButtonClassName}
```
azasypkin: Same here, can we use ```button``` instead?

app/js/views/themes-new.jsx
```diff
@@ -0,0 +1,221 @@
...
+            {serviceSelector}
+            {propertySelector}
+
+            <h2 className={headerClassName}>Do</h2>
+            {actionServiceSelector}
+            {actionPropertySelector}
+          </div>
+        </section>
+        <footer className="app-view__footer">
+          <NavigationMenu />
```
azasypkin: Either this one or pr#26 will need to be rebased to include ```foxbox``` property here :)

app/css/views/themes.css
```diff
@@ -0,0 +1,43 @@
+.themes__new-link {
+  padding-top: .5rem;
+  height: 3rem;
+}
+
+.themes-list {
```
azasypkin: Honestly all this top/bottom/line-height/absolute alignment look a bit messy, why not to use only left-right margins for nested elements if needed + ```align-items: center``` + top/bottom padding (if needed) on the ```li``` and ```flex: 1``` for ```themes-list__name```. Or I'm missing something?

app/css/views/themes-new.css
```diff
@@ -0,0 +1,60 @@
+.new-theme {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+  align-items: stretch;
+  position: absolute;
```
azasypkin: Why not just ```height: 100%``` instead of absolute positioning?

app/css/views/themes-new.css
```diff
@@ -0,0 +1,60 @@
...
+  right: 0;
+  padding: 0 4rem;
+  background: #ceedec;
+}
+
+.new-theme__header {
+  padding: 0;
+}
+
+.new-theme__header.new-theme__header--hidden {
```
azasypkin: Mmm why not just ```.new-theme__header--hidden``` here and in few places below?

azasypkin:
@gmarty I've left some questions/comments/nits, thanks!

app/js/controllers/themes.js
gmarty: My plan is to use the default case as the route for viewing a specific recipe like:
```
#themes/recipe-id
```

Would that be better to default the action parameter to `list` and rename the case in the switch?

app/js/views/themes-new.jsx
gmarty: Looks really good. Let's use that (later).

app/css/views/themes-new.css
gmarty: I know it sounds lame but I haven't been able to make `height: 100%` work. If you have any idea, please let me know.

app/js/views/themes-new.jsx
gmarty: Yes I agree that it looks ugly, but maybe we can postpone this refactoring for later?

gmarty:
Most of the comments and nits are addressed. Others will be implemented later.

PR #31, opened 2016-03-25
Use Taxonomy based Service API. r=gmarty

azasypkin:
r? @gmarty If something is not critical I can fix it in the follow ups :)

### DIFF
--- a/app/js/lib/foxbox.js
+++ b/app/js/lib/foxbox.js
@@ -40,51 +40,85 @@ let themes = [
 
 const _private = {
   isPollingEnabled: Symbol('isPollingEnabled'),
-  nextPollTimeout: Symbol('nextPollTimeout')
+  nextPollTimeout: Symbol('nextPollTimeout'),
+
+  fetchServices: Symbol('fetchServices')
 };
 
 /**
- * Request a JSON from a specified URL.
+ * Request a content of the specified type from a specified URL.
  *
  * @param {string} url The URL to send the request to.
- * @param {string} method The HTTP method (defaults to "GET").
- * @param {!Object} body An object of key/value.
+ * @param {string} contentType The content mime type (eg. image/jpeg).
+ * @param {string} [method=GET] The HTTP method (defaults to "GET").
+ * @param {Object} [body] An object of key/value.
  * @return {Promise}
  */
-const fetchJSON = function(url, method = 'GET', body = undefined) {
+const fetchContent = function(
+  url, contentType, method = 'GET', body = undefined
+) {
   method = method.toUpperCase();
 
   const req = {
     method,
-    headers: {
-      'Accept': 'application/json'
-    },
+    headers: { Accept: contentType },
     cache: 'no-store'
   };
 
   if (method === 'POST' || method === 'PUT') {
     req.headers['Content-Type'] = 'application/json;charset=UTF-8';
   }
+
   if (settings.session) {
     // The user is logged in, we authenticate the request.
     req.headers.Authorization = `Bearer ${settings.session}`;
   }
 
-  if (method === 'GET' || method === 'HEAD') {
-    body = undefined;
-  }
   if (body !== undefined) {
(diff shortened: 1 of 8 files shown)


app/js/lib/foxbox.js
```diff
@@ -40,54 +40,82 @@ let themes = [
...
 /**
- * Request a JSON from a specified URL.
+ * Request a content of the specified type from a specified URL.
  *
- * @param {string} url The URL to send the request to.
- * @param {string} method The HTTP method (defaults to "GET").
- * @param {!Object} body An object of key/value.
+ * @param {String} url The URL to send the request to.
+ * @param {String} contentType The content mime type (eg. image/jpeg).
+ * @param {String} [method=GET] The HTTP method (defaults to "GET").
```
gmarty: I usually use the annotations types of [Closure Compiler](https://developers.google.com/closure/compiler/docs/js-for-compiler#types) where the type `@param {string}` is lower case (all types are lower case except `Object` and `Array`).

app/js/lib/foxbox.js
```diff
@@ -40,54 +40,82 @@ let themes = [
...
   }
+
   if (settings.session) {
     // The user is logged in, we authenticate the request.
     req.headers.Authorization = `Bearer ${settings.session}`;
   }
 
-  if (method === 'GET' || method === 'HEAD') {
-    body = undefined;
-  }
```
gmarty: Why removing that? AFAIK it's part of the fetch API.

app/js/lib/foxbox.js
```diff
@@ -40,54 +40,82 @@ let themes = [
...
+        throw new TypeError(
+          `The response returned a ${res.status} HTTP status code.`
+        );
       }
 
-      throw new TypeError(`The response returned a ${res.status} HTTP status code.`);
+      return res;
+    }).catch((e) => {
+      console.error('Error occurred while fetching content: ', e);
+      throw e;
```
gmarty: I don't think we should catch here, otherwise the application won't have any ways to know whether the request was successful or not.
It's up to the app to append `catch()` to the promise returned by the request.

app/js/lib/foxbox.js
```diff
@@ -425,7 +453,7 @@ export default class Foxbox extends Service {
               .then((state) => service.state = state);
           })
         ).then(() => services);
-      });
+      });*/
```
gmarty: Are we going to have state services in the future? If no, let's remove that bit rather than just commenting out.

app/js/lib/foxbox.js
```diff
@@ -40,54 +40,82 @@ let themes = [
...
-        return res.json();
+      if (!res.ok) {
+        throw new TypeError(
+          `The response returned a ${res.status} HTTP status code.`
+        );
       }
 
-      throw new TypeError(`The response returned a ${res.status} HTTP status code.`);
+      return res;
+    }).catch((e) => {
```
gmarty: Line break before `.catch` please :-)

app/js/lib/foxbox.js
azasypkin: I'm not completely sure, but you're right we can get rid of it and return it back once/if we need it.

app/js/lib/foxbox.js
azasypkin: Ah, I forgot :)

app/js/lib/foxbox.js
azasypkin: Ah okay, I'm fine with that too - now we have a convention :)

app/js/views/services-list.jsx
```diff
@@ -21,4 +25,13 @@ export default class ServicesList extends React.Component {
       <ul className="service-list">{serviceNodes}</ul>
     );
   }
+
+  isUserFacingService(service) {
```
gmarty: This method should be moved to the `foxbox` lib and be called under the hood when you call `foxbox.getServices()`.

PR #34, opened 2016-03-25
Refactor the network related code out of the main class. r=azasypkin

gmarty:


app/js/lib/foxbox/foxbox.js
```diff
@@ -38,103 +28,20 @@ let themes = [
   }
 ];
 
-const _private = {
+// Private members.
+const p = {
```
azasypkin: Heh, you don't ike your ```_private``` anymore :D

app/js/lib/foxbox/foxbox.js
```diff
@@ -306,7 +187,7 @@ export default class Foxbox extends Service {
     if (searchParams.has('session_token')) {
       // There is a session token in the URL, let's remember it.
       // @todo Find a better way to handle URL escape.
-      settings.session = searchParams.get('session_token').replace(/ /g, '+');
+      this[p.settings].session = searchParams.get('session_token').replace(/ /g, '+');
```
azasypkin: __nit:__ too long line, let's keep it below 80.

app/js/lib/foxbox/foxbox.js
```diff
@@ -447,7 +278,7 @@ export default class Foxbox extends Service {
       return Promise.resolve();
     }
 
-    /*const fetchedServicesPromise = fetchJSON(`${this.origin}/services/list`)
+    /*const fetchedServicesPromise = this[p.net].fetchJSON(`${this[p.net].origin}/services/list`)
```
azasypkin: __nit:__ same long line remark here as well

app/js/lib/foxbox/foxbox.js
```diff
@@ -537,7 +369,7 @@ export default class Foxbox extends Service {
    */
   setServiceState(id, state) {
     return new Promise((resolve, reject) => {
-      fetchJSON(`${this.origin}/services/${id}/state`, 'PUT', state)
+      this[p.net].fetchJSON(`${this[p.net].origin}/services/${id}/state`, 'PUT', state)
```
azasypkin: __nit:__ too long line

app/js/lib/foxbox/foxbox.js
```diff
@@ -597,8 +429,8 @@ export default class Foxbox extends Service {
     let payload = { id: operation.id };
 
     if (operation.kind.typ === 'Binary') {
-      return fetchBlob(
-        `${this.origin}/api/v${settings.apiVersion}/channels/get`,
+      return this[p.net].fetchBlob(
+        `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/get`,
```
azasypkin: __nit:__ too long line

app/js/lib/foxbox/network.js
```diff
@@ -0,0 +1,206 @@
+'use strict';
+
+// Private members.
+const p = {
+  // Private properties.
+  // Whether we can connect to the box via a local connection.
```
azasypkin: __thinking aloud:__ :loud_sound: Honestly I'd put those comments in the constructor where actual variables are declared using these ```labels```, but up to you.

app/js/lib/foxbox/network.js
```diff
@@ -0,0 +1,206 @@
...
+
+    this[p.local] = false;
+    this[p.remote] = false;
+    this[p.pingInterval] = null;
+  }
+
+  /**
+   * Attach event listeners related to the connection status.
+   *
+   * @return {Promise}
```
azasypkin: __question:__ why do we return this useless promise here at all?

app/js/lib/foxbox/network.js
```diff
@@ -0,0 +1,206 @@
...
+        this.settings.onlineCheckingInterval);
+    }
+
+    this[p.pingBox]();
+
+    return Promise.resolve();
+  }
+
+  get origin() {
+    if (this[p.local]) {
```
azasypkin: __question:__ could you please give me a little more background here? We have two variables ```local``` and ```remote``` could they both be ```true```? If not, why do we need two such booleans instead of say ```isLocal``` and ```isConnected```?

app/js/lib/foxbox/network.js
```diff
@@ -0,0 +1,206 @@
...
+    } else if (this[p.remote]) {
+      return this.tunnelOrigin;
+    }
+
+    console.error('The box is out of reach.');
+    return this.localOrigin;
+  }
+
+  get localOrigin() {
+    return `${this.settings.localScheme}://${this.settings.localHostname}:${this.settings.localPort}`;
```
azasypkin: __nit:__ too long line here and below.

app/js/lib/foxbox/network.js
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  /**
+   * Request a content of the specified type from a specified URL.
+   *
+   * @todo Detect if the URL is relative, if so prepend this.origin.
+   *
+   * @param {string} url The URL to send the request to.
+   * @param {string} accept The content mime type (eg. image/jpeg).
+   * @param {string=} method The HTTP method (defaults to "GET").
```
azasypkin: You don't like ```default value```[1] in jsdoc for some reason (I see you got rid of it)? :) 

[1] http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

app/js/lib/foxbox/network.js
```diff
@@ -0,0 +1,206 @@
...
+   * Request a JSON from a specified URL.
+   *
+   * @param {string} url The URL to send the request to.
+   * @param {string} method The HTTP method (defaults to "GET").
+   * @param {Object} body An object of key/value.
+   * @return {Promise}
+   */
+  fetchJSON(url, method = 'GET', body = undefined) {
+    return this[p.fetch](url, 'application/json', method, body)
+      .then(response => response.json());
```
azasypkin: __Code Style Proposal:__ I'd propose always use ```() => {} and (a, b, c) => {}``` arrow function notation so that we always use the same style for all types of arrow functions and no matter if they have one, multiple or no parameters at all. IMO it's a bit easier to read, but it's subjective thing. What do you think?

app/js/lib/foxbox/network.js
```diff
@@ -0,0 +1,206 @@
...
+      });
+  }
+
+  /**
+   * Ping the box to detect whether we connect locally or remotely.
+   *
+   * @private
+   */
+  [p.pingBox]() {
+    this[p.ping](`${this.localOrigin}/ping`)
```
azasypkin: This method feels somewhat racy, but we have the same in master, so I'm fine with it in this pr :)

azasypkin:
@gmarty Looks really good! Just some questions and nits :)

app/js/lib/foxbox/foxbox.js
gmarty: I thought it was too verbose, so I shorten it.
But I really like this pattern and I'm reusing it in other parts of the lib.

azasypkin: Yep, let try this! :)

app/js/lib/foxbox/foxbox.js
gmarty: I agree.

In the future, I want `this[p.net].fetchJSON` et al. to be clever enough to detect if the URL passed is absolute or relative. If it's relative then it can automatically prepend `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/` so that we can focus more on the calls to the REST API rather than on those verbose details.

app/js/lib/foxbox/foxbox.js
gmarty: BTW, the line lengths limit is already enforced in jshint, but because of the transpilation to es5, the generated lines can be longer than the ones in the initial es6 code. So this line length enforcement ends up being rather useless.

Maybe we need to update the workflow to enforce coding style like that before the transpilation. I don't know if jshint or eshint supports the es6 syntax though.

app/js/lib/foxbox/network.js
gmarty: I was thinking it's good to have the same signature for all `init()` functions in the lib, but I agree it's pretty useless here.

app/js/lib/foxbox/network.js
gmarty: Because you can be not local and not remote (= disconnected). And if they're both `true`, then you can connect using either the local or remote connection (local is preferred though).
`local` can be understood as 'I can connect to the box using the local network' and `remote` means 'I can connect to the box using the internet'. They're not mutually exclusive.
I thought this implementation was easier when I worked on the logic for `pingBox()`.

app/js/lib/foxbox/network.js
gmarty: Yes, I've started using the [annotations system from Closure Compiler](https://developers.google.com/closure/compiler/docs/js-for-compiler#multiple-template-types) that doesn't support mentioning the default value. So I think it's redundant to use this type of annotation and an explanation in plain English (like `(defaults to "GET")`). So I decided to only keep the latter.

What do you think?

The rationale from using the annotations of Closure Compiler is that we're going to use it in our build system one day.
Also, in the future when we generate documentation from the code, we may want to add mentioning the default value of parameters if supported by the doc generator and if it doesn't cause any issues with Closure Compiler.

app/js/lib/foxbox/network.js
gmarty: Not sure I understand what you mean. You suggest:
```js
.then(response => {
  return response.json();
});
```
or
```js
.then((response) => response.json());
```

azasypkin: This one (always use round brackets around arrow function parameters):
```
.then((response) => response.json());
```

app/js/lib/foxbox/network.js
gmarty: Yes, the idea is we try both and see which one responds first.
It's better than wrapping both promises in a `Promise.all()`.

app/js/lib/foxbox/network.js
gmarty: I tend to think that the extra parens are useless. I like syntaxes that use less code as possible.

But we had this discussion on fxos mailing list a few months ago and some people seem to like it, so I'd reluctantly integrate this in the coding style :-P

app/js/lib/foxbox/network.js
azasypkin: Ah, I see with ```local is preferred``` it makes sense, thanks for the explanation.

app/js/lib/foxbox/network.js
azasypkin: > But we had this discussion on fxos mailing list a few months ago and some people seem to like it,

I believe it was only you and Kevin :P But on a serious note, that's totally fine, let's keep it like you want if we can't agree on new changes :)

app/js/lib/foxbox/network.js
azasypkin: Well, my main rationale is that JSDoc is tool/compiler/framework agnostic while Google Closure annotations are not. And even if start using Google Closure Compiler, it should not affect the way we document code, JSDoc and document generation tool based on it evolutionize independently and rapidly :)

app/js/lib/foxbox/network.js
azasypkin: It's not related to the PR, but should the previous statement ```local is preferred``` be replaced with ```the one that faster responds to ping is preferred```?  It may be important for any high-bandwidth requests/streams.

gmarty: I don't a think remote connection can ever be faster than local network. Am I wrong?
Also remote connection may result in data cost on mobile, while the local one is probably free.

azasypkin: > I don't a think remote connection can ever be faster than local network. Am I wrong?

I think it can happen (admit that very rarely though), just any local network glitch and remote will be faster :) But anyway it's not related to this patch

gmarty:
I fixed the issues mentioned except the ones related to coding styles and annotations until we agree on a consensus.

PR #38, opened 2016-03-31
Add badges and license

JohanLorenzo:
I found out we didn't have a direct link to travis. I copied the badges from fxfox/foxbox. The coverage one is missing because coverage will require more investigation. 

r? @gmarty

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,373 @@
+Mozilla Public License Version 2.0
+==================================
+
+1. Definitions
+--------------
+
+1.1. "Contributor"
+    means each individual or legal entity that creates, contributes to
+    the creation of, or owns Covered Software.
+
+1.2. "Contributor Version"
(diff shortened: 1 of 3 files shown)


gmarty:
You need to update the license type in `package.json` too (the key is `"license"`).

JohanLorenzo:
~~r? @gmarty~~ . I have a permafail on Travis. I currently don't see what could cause it. I also don't manage to repro.

I'll take a closer look at it, but I don't wanna waste your time, in the meantime

gmarty:
In the latest master, @azasypkin fixed a race condition that affected the integration tests. See https://github.com/fxbox/app/pull/43#discussion_r58229896 for details.
Maybe your failure here is related to that.

JohanLorenzo:
That seems to be it. Patch rebased r? @gm

PR #39, opened 2016-03-31
Use ESLint for JS and JSX (fixes #37). r=gmarty

### LINKED ISSUE #37: Use eslint for linting code
Let's switch to [eslint](http://eslint.org/) that supports JS and JSX out of the box!

azasypkin:


.eslintrc
```diff
@@ -0,0 +1,41 @@
...
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module",
+    "ecmaFeatures": {
+      "impliedStrict": true,
+      "jsx": true
+    }
+  },
+
+  "rules": {
```
azasypkin: I've migrated most of the rules from ```.jshintrc```

app/js/lib/foxbox.js
```diff
@@ -16,8 +16,6 @@ let _tunnelConfigured = false;
 let _local = false;
 // Whether we can connect to the box via a remote connection.
 let _remote = false;
-// A reference to the interval to get the online status.
-let _onlineInterval = null;
```
azasypkin: It's not used and ESLint is complaining, if we decide to clear interval we can return it back

app/js/views/modal.jsx
```diff
@@ -21,3 +21,10 @@ export default class Modal extends React.Component {
     );
   }
 }
+
+Modal.propTypes = {
```
azasypkin: Here are changes suggested by eslint react plugin, I actually like them. See [1] and [2].

[1] https://facebook.github.io/react/docs/reusable-components.html
[2] https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md

gulpfile.js
```diff
@@ -142,8 +142,10 @@ gulp.task('travis', ['lint', 'loader-polyfill', 'babel-app']);
  * Build the app.
  */
 gulp.task('build', function(cb) {
-  runSequence(['clobber-app'], ['loader-polyfill', 'copy-app'], ['babel-app'],
-    ['lint'], ['remove-useless'], cb);
+  runSequence(
+    ['lint'], ['clobber-app'], ['loader-polyfill', 'copy-app'], ['babel-app'],
```
azasypkin: I believe linter should be the first one here to not waste time if something is wrong in the code

.eslintrc
```diff
@@ -0,0 +1,41 @@
...
+    "comma-style": ["error", "last"],
+    "curly": ["error"],
+    "max-len": ["error", { "code": 80, "tabWidth": 2 }],
+    "new-cap": ["error"],
+    "no-alert": ["warn"],
```
azasypkin: It should eventually be ```error```, not ```warn``` - but we need to get rid of ```prompt``` in tag view.

gulpfile.js
gmarty: That's a very good idea!

app/js/lib/foxbox/foxbox.js
```diff
@@ -355,7 +355,7 @@ export default class Foxbox extends Service {
     return this[p.net].fetchJSON(`${this[p.net].origin}/services/${id}/state`)
       .then(res => {
         if (!res) {
-          throw new Error(`The action couldn't be performed.`);
+          throw new Error('The action couldn\'t be performed.');
```
gmarty: Wait, eslint don't support back tick as a regular string delimiter? That's a major annoyance to force using single quote. Can't we ignore these warning for now and maybe later write our own plugin for that?

.eslintrc
```diff
@@ -0,0 +1,41 @@
+{
+  "parser": "babel-eslint",
+
+  "parserOptions": {
+    "ecmaVersion": 6,
```
gmarty: Shame that eslint es version doesn't follow the new naming like `es2015` and `es2016`...

app/js/lib/foxbox/foxbox.js
azasypkin: I believe if we don't use parameters inside ```template string``` it complains about it - that's good IMO, cause it's just string no need in anything else. If we have parameters then using ```template string``` is good.

Does it make sense?

.eslintrc
azasypkin: :)

app/js/lib/foxbox/foxbox.js
gmarty: It's a shame because single and double quotes are frequently found in static strings.
I've always wanted to use a coding style that enforces the use of back tick as a string delimiter regardless of the presence of a variable inside the string.
But I understand your point so let's stick with single quote and back tick only if there's a parameter.

azasypkin: Although it seems we can use ```quotes: ["error", "single", "backtick"]``` to enable backticks for regular strings, I have a mixed feelings about using backticks for ordinary strings, maybe it's just me

app/js/lib/foxbox/settings.js
```diff
@@ -33,12 +33,19 @@ const storage = localStorage ? localStorage : {
 
 export default class Settings extends Model {
   constructor() {
+    let localHostName = storage.getItem(`${PREFIX}localHostname`) ||
+      DEFAULT_HOSTNAME;
+
+    let pollingEnabled = storage.getItem(`${PREFIX}pollingEnabled`) !== null ?
```
gmarty: you can probably safely use `const` here as the scope is limited to the `constructor()` method.

app/js/lib/foxbox/db.js
```diff
@@ -62,7 +62,7 @@ export default class Db {
     return new Promise((resolve, reject) => {
       this.db.close();
 
-      var req = indexedDB.deleteDatabase(DB_NAME);
+      let req = indexedDB.deleteDatabase(DB_NAME);
```
gmarty: Comment: Warning on `var` is IMO the best feature of eslint in es6 mode!

app/js/lib/foxbox/foxbox.js
azasypkin: > But I understand your point so let's stick with single quote and back tick only if there's a parameter.

Sounds good! We can return back to this style later if we feel we need this change :)

app/js/lib/foxbox/db.js
azasypkin: Heh, yeah!

app/js/lib/foxbox/settings.js
azasypkin: Right, will change

.eslintrc
```diff
@@ -0,0 +1,41 @@
+{
+  "parser": "babel-eslint",
```
gmarty: Can you highlight the benefit or using the babel parser?

azasypkin: Not really, I just use it cause we use Babel for transpilation, didn't investigate real benefits. Do you prefer to return back to default Esprima?

app/js/views/modal.jsx
gmarty: I agree. I like the use of `propTypes` because it can be useful to all sort of static analysis tools.
I just hope that won't be too much troubles when creating/changing components.
Also it may add some useless code that's not really needed on the production code (The prop validation only happens on development mode of React). If the `propTypes` property of components are left as is by Babel/React maybe we can write a gulp task to remove them from the transpiled code.

.eslintrc
gmarty: No, I was just curious to know if it was required when working with Babel. Let's keep it.

gmarty:
That's an amazing job Oleg! Congratulations! r+

tests/unit/test-main.js
```diff
@@ -1,3 +1,5 @@
+/*eslint no-var: "off"*/
```
gmarty: nit: I think comments are more readable when wrapped in spaces:
```js
/* eslint no-var: "off" */
```

app/js/views/modal.jsx
azasypkin: > I just hope that won't be too much troubles when creating/changing components.

Right, let's see how it goes.

> If the propTypes property of components are left as is by Babel/React maybe we can write a gulp task to remove them from the transpiled code.

Good point, we need to check, hopefully ```react``` babel preset will take care of this, but I'm not sure.

tests/unit/test-main.js
azasypkin: True! :)

azasypkin:
Nits fixed, Travis is green, thanks for review!

PR #41, opened 2016-04-01
 Update foxbox and camera service view to be compatible with the latest taxonomy (fixes #40). r=gmarty

### LINKED ISSUE #40: Camera view is broken by incompatible API change
See https://github.com/fxbox/foxbox/pull/290

+ We need to somehow notify about breaking changes in API.

azasypkin:
Depends on https://github.com/fxbox/foxbox/pull/296 and https://github.com/fxbox/app/pull/39 and https://github.com/fxbox/app/pull/43

### DIFF
--- a/app/js/lib/foxbox/foxbox.js
+++ b/app/js/lib/foxbox/foxbox.js
@@ -38,7 +38,8 @@ const p = {
   nextPollTimeout: Symbol('nextPollTimeout'),
 
   // Private methods.
-  fetchServices: Symbol('fetchServices')
+  fetchServices: Symbol('fetchServices'),
+  getOperationValueType: Symbol('getOperationValueType')
 };
 
 /**
@@ -422,11 +423,12 @@ export default class Foxbox extends Service {
   }
 
   performSetOperation(operation, value) {
+    let operationType = this[p.getOperationValueType](operation.kind);
     return this[p.net].fetchJSON(
       `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/set`,
       'PUT',
       // Query operation by id.
-      [[{ id: operation.id }, { [operation.kind.typ]: value }]]
+      [[{ id: operation.id }, { [operationType]: value }]]
     );
   }
 
@@ -465,5 +467,31 @@ export default class Foxbox extends Service {
           };
         });
       });
+  }
+
+  /**
+   * Returns value type string for the specified operation kind.
+   *
+   * @param {string|Object} operationKind Kind of the operation, string for the
+   * well known type and object for the Extension channel kind.
+   * @return {string}
+   * @private
+   */
+  [p.getOperationValueType](operationKind) {
+    if (!operationKind) {
+      throw new Error('Operation kind is not defined!');
+    }
+
+    // Operation kind can be either object or string.
+    if (typeof operationKind === 'object') {
+      return operationKind.typ;
+    }
+
+    switch (operationKind) {
+      case 'TakeSnapshot':
+        return 'Unit';
+      default:
+        return operationKind;
+    }
   }
 }
--- a/app/js/views/services/camera.jsx
+++ b/app/js/views/services/camera.jsx
@@ -15,28 +15,63 @@ export default class CameraService extends React.Component {
     };
   }
 
+  /**
+   * Gets service operation with the specified alias.
+   * 
+   * @param {Array<Object>} operations List of available service operations.
+   * @param {string} alias Alias of the operation we're looking for.
+   * @return {Object} Operation associated with the specified alias.
+   *
+   * @private
+   */
   getOperationByAlias(operations, alias) {
     let operationKey = Object.keys(operations).find((key) => {
       let operation = operations[key];
-      return operation.kind && operation.kind.kind == alias;
+
+      if (typeof operation.kind === 'object') {
+        return operation.kind.kind === alias;
+      }
+
+      return operation.kind === alias;
     });
 
     return operations[operationKey];
   }
 
+  /**
+   * Gets service "set" operation with the specified alias.
+   *
+   * @param {string} alias Alias of the operation we're looking for.
+   * @return {Object} Operation associated with the specified alias.
+   *
+   * @private
+   */
   getSetOperation(alias) {
     return this.getOperationByAlias(this.service.setters, alias);
   }
 
+  /**
+   * Gets service "get" operation with the specified alias.
+   *
+   * @param {string} alias Alias of the operation we're looking for.
+   * @return {Object} Operation associated with the specified alias.
+   *
+   * @private
+   */
   getGetOperation(alias) {
     return this.getOperationByAlias(this.service.getters, alias);
   }
 
+  /**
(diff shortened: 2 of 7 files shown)


azasypkin:
Okay, all dependencies are landed, time to finish first react component test :)

app/js/lib/foxbox/foxbox.js
```diff
@@ -466,4 +468,29 @@ export default class Foxbox extends Service {
         });
       });
   }
+
+  /**
```
azasypkin: Trying to not spend too much time to make api as generic as possible as it's going to change again soon.

tests/unit/views/services/camera_test.js
```diff
@@ -0,0 +1,112 @@
+import Rx from 'rxjs';
+import React from 'components/react';
+
+import CameraServiceView from 'js/views/services/camera';
+
+// See capabilities at https://facebook.github.io/react/docs/test-utils.html
+const TestUtils = React.addons.TestUtils;
+
+function waitFor(fn) {
```
azasypkin: This is an experiment, not sure if you like it :) Here I just want to play with the way to test async actions where we don't have access to the underlying promises to wait on (e.g a bunch of async actions in DOM event handler). Idea is exactly the same as ```waitFor``` from MarionetteJS and partially borrowed from [1].

Tell me what you think!

P.S. Such utility methods should be moved to a separate file, but I like the rule - ```Do it for yourself, if you/someone-else need(s) in another file - only then make it generic``` :)

[1] http://gregbabiars.com/acceptance-testing-a-react-app-using-react-test-utils-pretender-and-rxjs/

tests/unit/views/services/camera_test.js
```diff
@@ -0,0 +1,112 @@
...
+        name: 'service-name'
+      }
+    });
+
+    component = TestUtils.renderIntoDocument(
+      <CameraServiceView service={serviceStub} foxbox={foxboxStub} />
+    );
+  });
+
+  it('Renders itself in the correct state', function () {
```
azasypkin: Simple tests, just to have example of testing React components. Still hesitating about too comprehensive tests on this stage + API is being actively developed/changed.

app/js/lib/foxbox/foxbox.js
```diff
@@ -466,4 +468,29 @@ export default class Foxbox extends Service {
...
+   * Returns value type string for the specified operation kind.
+   * @param {string|Object} operationKind Kind of the operation, string for the
+   * well known type and object for the Extension channel kind.
+   * @returns {string}
```
gmarty: I usually use `@return` instead of `@returns` which is supported by Closure Compiler.

app/js/views/services/camera.jsx
```diff
@@ -15,28 +15,60 @@ export default class CameraService extends React.Component {
     };
   }
 
+  /**
+   * Gets service operation with the specified alias.
```
gmarty: nit: I usually add a line break between the description and the JSDoc annotations. Then all the annotation are grouped together. What do you think?

tests/unit/views/services/camera_test.js
gmarty: I think I like it and it works well when you're used to work with MarionetteJS.
I also think it should be moved to a separate file, but we can also wait until we need it somewhere else.

tests/unit/views/services/camera_test.js
```diff
@@ -0,0 +1,112 @@
...
+
+        sinon.assert.calledWith(
+          foxboxStub.performGetOperation, serviceStub.getters['getter-id']
+        );
+
+        assert.match(component.refs.snapshotPreview.src, /^blob:/);
+      })
+      .then(done, done);
+  });
+});
```
gmarty: So we still write unit and integration tests in es5. If I understand correctly all the environments we're running the tests in support es6.
What do you think we should do? For now, I'm in support of continuing writing them in es5 for a better compatibility when we want to support more browsers. But we may inadvertently use some es6 features that we're used to (for example `const` in this file).
In not such a distant future we may want to use `await` for the tests to improve readability. So let's start thinking forward. Shall we start writing tests in es6 now and defer transpilation to the moment we actually need it? Or shall we have test transpilation in place now? Or shall we continue writing test in es5?

gmarty:
r+
I left a few nits and started a discussion about writing tests in es5 or es2015.

tests/unit/views/services/camera_test.js
azasypkin: > So we still write unit and integration tests in es5.

Mm, why do you think we write __unit tests__ in es5? Now we write them in the same way as we write the app - in es6 (const, imports, arrows) and then it's transpiled to be run in es5 since app is transpiled as well and we need to somehow bridge them. Having common ground between app and tests brings less inconsistencies.

Regarding integration tests - I'd do the same for them, but I have no idea about their current state/environment and etc :)

> What do you think we should do? For now, I'm in support of continuing writing them in es5 for a better compatibility when we want to support more browsers. But we may inadvertently use some es6 features that we're used to (for example const in this file).

I still think we should write tests in the same way we write app (es6, ES2015/16, no mental switch between app and tests), we should run tests in the same way we run app (transpiled to es5, run tests using the same files that are run on production) :)

> In not such a distant future we may want to use await for the tests to improve readability. So let's start thinking forward. Shall we start writing tests in es6 now and defer transpilation to the moment we actually need it? Or shall we have test transpilation in place now? Or shall we continue writing test in es5?

As I said above, I'd keep both tests and app workflows (coding and running) in sync. Regarding ```async\await``` - I think we can use it already with Babel transpilation... Or maybe I didn't get your idea?

What do you think?

app/js/lib/foxbox/foxbox.js
azasypkin: Sounds good to me

app/js/views/services/camera.jsx
azasypkin: Yep, I think we can even make it a style guide requirement :)

tests/unit/views/services/camera_test.js
azasypkin: Good, let's experiment with it then!

tests/unit/views/services/camera_test.js
gmarty: That makes sense, let's not write too much code that'll be throw away soon.
We'll have to implement the new design at some point and I expect it'll change a lot of things.

azasypkin:
> r+
I left a few nits and started a discussion about writing tests in es5 or es2015.

Nits fixed, replied to your question regarding ES2015 and tests, let's discuss further to have common ground and goal!

Travis is green.

tests/unit/views/services/camera_test.js
gmarty: Nice. I haven't realised the tests where transpiled too.
As you said that's good to avoid mental switch. But in this case there's many features we don't use (arrow functions, const, let...).

Regarding `await` it doesn't ship with the `es2015` preset of Babel. It comes as [a plugin](https://babeljs.io/docs/plugins/transform-async-to-generator/). So we just need to add it when we want to use it.

PR #43, opened 2016-04-01
Get rid of Bower dependencies in favor of NPM ones (fixes #42). r=gmarty

### LINKED ISSUE #42: Get rid of Bower dependencies in favor or NPM ones
Depends on https://github.com/fxbox/app/issues/37

azasypkin:
Depends on https://github.com/fxbox/app/pull/39


package.json
```diff
@@ -20,6 +20,9 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "react": "^0.14.8",
+    "react-dom": "^0.14.8",
+    "fxos-mvc": "fxos-eng/mvc#348b3ae",
```
gmarty: I think we can safely rely on the master for `fxos-mvc`.

azasypkin: Okay, likely we're only users of it anyway :)

gulpfile.js
```diff
@@ -9,15 +9,14 @@
 
 var gulp = require('gulp');
 
-var buildModules = __dirname + '/node_modules/fxos-build/node_modules/';
```
azasypkin: Since we have our own bootstrap.js, I don't see why we need ```fxos-build``` at all - it looks like just an intermediate layer between our app and modules we need, but with more hassle and stale versions. app polyfil is just alameda (v0.2.0), I just use the latest alameda version directly (v1.0.0). And it will be much easier to update modules we use directly.

What do you think?

karma.conf.js
```diff
@@ -4,19 +4,12 @@ module.exports = function(config) {
...
     basePath: 'dist/app',
 
-    frameworks: ['mocha', 'chai', 'chai-as-promised', 'chai-sinon'],
+    frameworks: [
+      'mocha', 'requirejs', 'chai', 'chai-as-promised', 'chai-sinon'
+    ],
 
     // List of files / patterns to load in the browser.
     files: [
-      'components/react/react.js',
```
azasypkin: with React via ```import``` I can now get rid of this nasty piece of code :)

gulpfile.js
gmarty: That works for me. `fxos-build` has been super helpful at the beginning to start with a working development workflow. But now we need something more custom and as we grow out of the prototype phase we can dedicate more time to a solid build system.

karma.conf.js
gmarty: Tadaaaaa!

package.json
```diff
@@ -20,20 +20,29 @@
...
+    "fxos-mvc": "fxos-eng/mvc#master",
     "babel-plugin-transform-es2015-modules-amd": "^6.6.5",
     "babel-preset-es2015": "^6.6.0",
     "babel-preset-react": "^6.5.0",
-    "fxos-build": "^0.8.0",
-    "gulp": "^3.9.1",
-    "gulp-rename": "^1.2.2"
+    "alameda": "^1.0.0",
+    "gulp": "^3.9.1"
   },
```
gmarty: Did you install the dependencies with `npm install xxx --save`? If so, it should sort the dependencies in alphabetic order.

azasypkin: Mmm, nope, I usually add new package to the package list and run ```npm install``` and it installs everything I've just added

app/js/lib/foxbox/foxbox.js
```diff
@@ -67,6 +67,8 @@ export default class Foxbox extends Service {
     this[p.settings] = new Settings();
     this[p.db] = new Db();
     this[p.net] = new Network(this[p.settings]);
+
+    this.boxes = [];
```
azasypkin: That's the tiny thing absence of which was making my life harder :) When app is run for integration test, there is a race, app tries to connect to the remote registration service and fails, then login screen fails (boxes is undefined). And only then test modifies local storage to switch to local registration service, but login screen is already broken :/

gmarty:
At the moment we use React in dev mode, which is good enough. But at some point in the future, we need to use the production builds of React (the `*.min.js` files, they're more compact and optimised). I know this is out of scope of this patch, but that'd be good if this PR could pave the way to that.

I suggest that the `deploy.sh` script swaps the react dev scripts with the production ones using a dedicated gulp task. It seems that this is something that this patch can make possible. Do you agree?

app/js/lib/foxbox/foxbox.js
gmarty: I'm OK with adding this change to this patch.
Also it's a good practice to always set the properties of a class to their default values in the `constructor()`. I wonder if eslint can help us with enforcing it as a coding standard.
BAD:
```js
class MyClass {
 constructor() {
 }

 method() {
  this.prop = value;
 }
}
```
GOOD:

```js
class MyClass {
 constructor() {
  this.prop = defaultValue;
 }

 method() {
  this.prop = value;
 }
}
```

What do you think?

azasypkin: I totally agree, it also clearly indicates what properties are defined for the particular class/instance. I've seen that there is a way to do this for react classes/components that likely can be linted with react eslint plugin. Maybe we can somehow enforce this for ordinary classes as well - need to dig.

package.json
gmarty: The finicky person is me is nit'ing on sorting dependencies in alphabetic order, but my rational side is trying to let it go with being finicky :-)
Do what works best for you.

azasypkin:
> I suggest that the deploy.sh script swaps the react dev scripts with the production ones using a dedicated gulp task. It seems that this is something that this patch can make possible. Do you agree?

Yep, I agree that we'll need this very soon actually. I also see this as a gulp task, I'm just holding this PR for now, until app structure stabilizes a bit more (I have few more slight changes in the upcoming unit test patch). My plan was to introduce "production" build in the patch that will split gulpfile into several modules ```build/bla1.js, build/bla2.js``` with gulpfile.js that imports them.

Does it sound good to you?

package.json
azasypkin: Ha-ha, I like when everything is sorted as well, I'll sort them :)

gmarty:
Yes, it seems we agree on everything.
Let's get this one landed! r+.

azasypkin:
package.json dependencies are sorted and Travis is happy, thanks for review!

PR #48, opened 2016-04-04
Add a class for the recipes. r=azasypkin

gmarty:
@azasypkin: This patch is a bit messy with a lot of code duplication and saving a rule to the box doesn't seem to work.
But if you agree, let's get it landed first and we can fix and refactor it. It's already big.

### DIFF
--- a/app/css/views/themes.css
+++ b/app/css/views/themes.css
@@ -42,4 +42,5 @@
   margin: 0 0 0 1rem;
   padding: 0;
   right: 0;
+  font-size: 0;
 }
--- a/app/js/lib/foxbox/recipes.js
+++ b/app/js/lib/foxbox/recipes.js
@@ -1,0 +1,299 @@
+'use strict';
+
+// @todo Proxy recipes in the db and emit events on change (see services).
+
+// Private members.
+const p = {
+  // Private properties.
+  settings: Symbol('settings'),
+  net: Symbol('net')
+};
+
+export default class Recipe {
+  constructor(props) {
+    this[p.settings] = props.settings;
+    this[p.net] = props.net;
+  }
+
+  /**
+   * Returns a promise resolving to a list of recipes.
+   *
+   * @return {Promise}
+   */
+  getAll() {
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
+      // Get all recipes.
+      .then(services => services.filter(service =>
+        service.adapter === 'thinkerbell-adapter' &&
+        service.id !== 'thinkerbell-root-service'
+      ))
+      // Fetch their respective enabled status.
+      .then(recipes => {
+        const promises = recipes.map(recipe => {
+          const payload = [
+            { id: `${recipe.id}/get_enabled` }
+          ];
+
+          return this[p.net].fetchJSON(
+            `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/` +
+            'channels/get',
+            'PUT',
+            payload
+          );
+        });
+
+        return Promise.all(promises)
+          // Map all the recipes to a more user-friendly format.
+          .then(servicesEnabled => recipes.map((recipe, index) => ({
+            id: recipe.id,
+            label: recipe.id,
+            enabled: servicesEnabled[index][`${recipe.id}/get_enabled`]
+              .OnOff === 'On'
+          })));
+      });
+  }
+
+  getServicesWithGetters() {
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
+      // Get all services but the recipes.
+      .then(services => services.filter(service =>
+        service.adapter != 'thinkerbell-adapter' &&
+        service.adapter != 'webpush@link.mozilla.org'
+      ))
+      // Keep only the services with getters.
+      .then(services => services.filter(service =>
+        Object.keys(service.getters).length
+      ))
+      // User friendly name.
+      .then(services => services.map(service => {
+          switch (service.id) {
+            case 'service:clock@link.mozilla.org':
+              service.name = 'Everyday';
+              break;
+
+            default:
+              service.name = service.id;
+              break;
+          }
+
+          return service;
+        }
+      ))
+      // User friendly getters.
+      .then(services => services.map(service => {
+          switch (service.id) {
+            case 'service:clock@link.mozilla.org':
+              service.getters = [
+                {
+                  label: 'in the morning',
+                  value: 'service:clock@link.mozilla.org,CurrentTimeOfDay,eq,' +
+                  '08:00am'
+                },
+                {
+                  label: 'in the afternoon',
+                  value: 'service:clock@link.mozilla.org,CurrentTimeOfDay,eq,' +
+                  '02:00pm'
+                },
+                {
+                  label: 'in the evening',
+                  value: 'service:clock@link.mozilla.org,CurrentTimeOfDay,eq,' +
+                  '06:00pm'
+                }
+              ];
+              break;
+
+            default:
+              //service.getters = service.getters;
+              break;
+          }
+
+          return service;
+        }
+      ))
+      // Map all the recipes to a more user-friendly format.
+      .then(services => services.map(service => ({
+        id: service.id,
+        label: service.name,
+        get: service.getters
+      })));
+  }
+
+  getServicesWithSetters() {
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
+      // Get all services but the recipes.
+      .then(services => services.filter(service =>
+        service.adapter != 'thinkerbell-adapter' &&
+        service.adapter != 'webpush@link.mozilla.org'
+      ))
+      // Keep only the services with getters.
+      .then(services => services.filter(service =>
+        Object.keys(service.setters).length
+      ))
+      // User friendly name.
+      .then(services => services.map(service => {
+          switch (service.id) {
+            case 'espeak@link.mozilla.org':
+              service.name = 'say';
+              break;
+
+            default:
+              service.name = service.id;
+              break;
+          }
+
+          return service;
+        }
+      ))
+      // User friendly getters.
+      .then(services => services.map(service => {
+          switch (service.id) {
+            case 'espeak@link.mozilla.org':
+              service.setters = [
+                {
+                  label: '"Good morning!"',
+                  value: 'espeak@link.mozilla.org,Sentence,String,' +
+                  '"Good morning!"'
+                },
+                {
+                  label: '"Good afternoon!"',
+                  value: 'espeak@link.mozilla.org,Sentence,String,' +
+                  '"Good afternoon!"'
+                },
+                {
+                  label: '"Good evening!"',
+                  value: 'espeak@link.mozilla.org,Sentence,String,' +
+                  '"Good evening!"'
+                }
+              ];
+              break;
+
+            default:
+              //service.setters = service.setters;
+              break;
+          }
+
+          return service;
+        }
+      ))
+      // Map all the recipes to a more user-friendly format.
+      .then(services => services.map(service => ({
+        id: service.id,
+        label: service.name,
+        set: service.setters
+      })));
+  }
+
+  /**
+   * Create a new recipe.
+   *
+   * @return {Promise}
+   */
+  add(recipe) {
+    const [idGet, kindGet, rangeKeyGet, rangeValGet] = recipe.getter.split(',');
+    const [idSet, kindSet, rangeKeySet, rangeValSet] = recipe.setter.split(',');
+
+    recipe = {
+      rules: [
+        {
+          conditions: [
+            {
+              source: [
+                {
+                  id: idGet
+                }
+              ],
+              kind: {
+                kind: kindGet
+              },
+              range: {
+                [rangeKeyGet]: rangeValGet
+              }
+            }
+          ],
+          execute: [
+            {
+              destination: [
+                {
+                  id: idSet
+                }
+              ],
+              kind: {
+                kind: kindSet
+              },
+              value: {
+                [rangeKeySet]: rangeValSet
+              }
+            }
+          ]
+        }
+      ]
+    };
+
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/set`,
+      'PUT',
+      [
+        [
+          [
+            { id: 'thinkerbell-add-rule' }
+          ],
+          {
+            ThinkerbellRule: {
+              name: 'foo',
+              source: JSON.stringify(recipe)
+            }
+          }
+        ]
+      ]
+    );
+  }
+
+  /**
+   * Remove a recipe with the associated id.
+   *
+   * @param {string} id
+   * @return {Promise}
+   */
+  remove(id) {
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/set`,
+      'PUT',
+      [
+        [
+          [
+            { id: `${id}/remove` }
+          ],
+          null
+        ]
+      ]
+    );
+  }
+
+  /**
+   * Enable or disable the specified recipe.
+   *
+   * @param {string} id
+   * @param {boolean=} value Whether to enable or disable. Enable by default.
+   * @return {Promise}
+   */
+  toggle(id, value = true) {
+    const textValue = value ? 'On' : 'Off';
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/set`,
+      'PUT',
+      [
+        [
+          [
+            { id: `${id}/set_enabled` }
+          ],
+          {
+            OnOff: textValue
+          }
+        ]
+      ]
+    );
+  }
+}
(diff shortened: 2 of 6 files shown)


app/css/views/themes.css
```diff
@@ -42,4 +42,5 @@
   margin: 0 0 0 1rem;
   padding: 0;
   right: 0;
+  font-size: 0;
```
azasypkin: __question:__ why do we need that and ```x``` as button text (what we really need there is ```title``` btw)?

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  /**
+   * Returns a promise resolving to a list of recipes.
+   *
+   * @return {Promise}
+   */
+  getAll() {
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
```
azasypkin: Not in this PR, but it feels that we need something like ```this[p.net].baseApiURL``` so that we can do ````${this[p.net].baseApiURL}\services````

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+        const promises = recipes.map(recipe => {
+          const payload = [
+            { id: `${recipe.id}/get_enabled` }
+          ];
+
+          return this[p.net].fetchJSON(
+            `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/` +
+            'channels/get',
+            'PUT',
+            payload);
```
azasypkin: __question:__ I've noticed we use different closing bracket style, I used to put it into a separate line (as curly braces, to visually enforce block boundaries and not to keep many braces in a single place) in case arguments are wrapped and you keep it together with the last argument. Let's agree on the same style?

```
callA(
  'aaaaaaaaaaaaaaaaa',
  'bbbbbbbbbbbbbbbbb',
  callC(
     'dddddddddddddddd',
     'eeeeeeeeeeeeeeee'
  )
);

vs

callA(
  'aaaaaaaaaaaaaaaaa',
  'bbbbbbbbbbbbbbbbb',
  callC(
    'dddddddddddddddd',
    'eeeeeeeeeeeeeeee'));
```

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+        });
+
+        return Promise.all(promises)
+          .then(servicesEnabled => {
```
azasypkin: __nit:__ mm what the purpose of this ```then``` branch? It feels redundant.

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+            return servicesEnabled;
+          })
+          // Map all the recipes to a more user-friendly format.
+          .then(servicesEnabled => recipes.map((recipe, id) => ({
```
azasypkin: __nit:__ Using ```id``` here feels confusing, it looks like ```index```

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+            id: recipe.id,
+            label: recipe.id,
+            enabled: servicesEnabled[id][`${recipe.id}/get_enabled`]
+              .OnOff === 'On'
+          })));
+      });
+  }
+
+  getServicesWithGetters() {
```
azasypkin: Hmm, shouldn't you query ```http://.../api/v1/channels/getters``` here instead?

Here is the response I get locally
```json
[{
    "adapter": "ip-camera@link.mozilla.org",
    "id": "getter:image_list.18db9b78-f188-11e5-9ce9-3ca9f47e346c@link.mozilla.org",
    "kind": {
        "adapter": "IPCam Adapter",
        "kind": "image_list",
        "typ": "Json",
        "vendor": "foxlink@mozilla.com"
    },
    "mechanism": "getter",
    "service": "service:18db9b78-f188-11e5-9ce9-3ca9f47e346c@link.mozilla.org",
    "tags": []
}, {
    "adapter": "ip-camera@link.mozilla.org",
    "id": "getter:image_newest.18db9b78-f188-11e5-9ce9-3ca9f47e346c@link.mozilla.org",
    "kind": {
        "adapter": "IPCam Adapter",
        "kind": "latest image",
        "typ": "Binary",
        "vendor": "foxlink@mozilla.com"
    },
    "mechanism": "getter",
    "service": "service:18db9b78-f188-11e5-9ce9-3ca9f47e346c@link.mozilla.org",
    "tags": []
}, {
    "adapter": "clock@link.mozilla.org",
    "id": "getter:timestamp.clock@link.mozilla.org",
    "kind": "CurrentTime",
    "mechanism": "getter",
    "service": "service:clock@link.mozilla.org",
    "tags": []
}, {
    "adapter": "webpush@link.mozilla.org",
    "id": "getter:subscription.webpush@link.mozilla.org",
    "kind": {
        "adapter": "WebPush adapter (built-in)",
        "kind": "WebPushSubscription",
        "typ": "Json",
        "vendor": "team@link.mozilla.org"
    },
    "mechanism": "getter",
    "service": "service:webpush@link.mozilla.org",
    "tags": []
}, {
    "adapter": "clock@link.mozilla.org",
    "id": "getter:timeofday.clock@link.mozilla.org",
    "kind": "CurrentTimeOfDay",
    "mechanism": "getter",
    "service": "service:clock@link.mozilla.org",
    "tags": []
}, {
    "adapter": "webpush@link.mozilla.org",
    "id": "getter:resource.webpush@link.mozilla.org",
    "kind": {
        "adapter": "WebPush adapter (built-in)",
        "kind": "WebPushResource",
        "typ": "Json",
        "vendor": "team@link.mozilla.org"
    },
    "mechanism": "getter",
    "service": "service:webpush@link.mozilla.org",
    "tags": []
}]
```

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+          return service;
+        }
+      ))
+      // User friendly getters.
+      .then(services => services.map(service => {
+          switch (service.id) {
+            case 'service:clock@link.mozilla.org':
+              service.getters = [
+                {
+                  label: 'in the morning',
```
azasypkin: I think all these user friendly labels should be done in the view itself, so that it can be easily localizable and to not pollute main foxbox lib, but I think we can leave it as is for this PR and improve in the follow ups.

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+          }
+
+          return service;
+        }
+      ))
+      // Map all the recipes to a more user-friendly format.
+      .then(services => services.map(service => ({
+        id: service.id,
+        label: service.name,
+        get: service.getters
```
azasypkin: why not ```getters: service.getters``` to stress that it's an array?

azasypkin: Same question about ```set``` below

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+      })));
+  }
+
+  /**
+   * Create a new recipe.
+   *
+   * @return {Promise}
+   */
+  add(recipe) {
+    console.log(recipe);
```
azasypkin: Looks like leftover from debugging :)

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+      // Map all the recipes to a more user-friendly format.
+      .then(services => services.map(service => ({
+        id: service.id,
+        label: service.name,
+        set: service.setters
+      })));
+  }
+
+  /**
+   * Create a new recipe.
```
azasypkin: I suspect JSDoc will be improved in the follow ups as well? :)

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+   *
+   * @return {Promise}
+   */
+  add(recipe) {
+    console.log(recipe);
+
+    const getter = recipe.getter.split(',');
```
azasypkin: __nit:__ ```let [id, kind, rangeKey, rangeValue] = recipe.getter.split(',');```?

Same for setters

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+  /**
+   * Remove a recipe with the associated id.
+   *
+   * @param {string} id
+   * @return {Promise}
+   */
+  remove(id) {
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/set`,
+      'PUT',
```
azasypkin: Hmm, Foxbox doesn't support ```DELETE``` method for recipe deletion?

app/js/lib/foxbox/recipes.js
```diff
@@ -0,0 +1,300 @@
...
+
+  /**
+   * Enable or disable the specified recipe.
+   *
+   * @param {string} id
+   * @param {boolean=} value Whether to enable or disable. Enable by default.
+   * @return {Promise}
+   */
+  toggle(id, value = true) {
+    const textValue = value ? 'On' : 'Off';
```
azasypkin: Foxbox on master complains about request format or something similar :/

app/js/views/themes.jsx
```diff
@@ -11,39 +14,15 @@ export default class Themes extends React.Component {
...
-
-  /**
-   * Delete a recipe.
-   *
-   * @param {number} id
-   */
-  handleOnDelete(id) {
-    this.foxbox.removeRecipe(id)
+  update() {
+    this.foxbox.recipes.getAll()
```
azasypkin: __nit:__ well we should either call "them" ```recipes``` or ```themes``` - it feels confusing to get themes when I request recipes :)

app/js/views/themes.jsx
```diff
@@ -51,24 +30,12 @@ export default class Themes extends React.Component {
...
-          <span className="themes-list__name">{theme.name}</span>
-          <button className="themes-list__remove"
-                  onClick={this.handleOnDelete.bind(this, id)}></button>
-        </li>
-      );
-    });
+    const themeItems = this.state.themes.map(theme => (
+      <ThemesListItem key={theme.id}
+                      theme={theme}
+                      update={this.update}
```
azasypkin: I'm not sure what's the recommended workflow for inter-component communication in React, but it feels natural for the ```ThemesListItem``` to expose event (```onDeleted```? I see it's only used for this right now) instead that we subscribe for in this component and do the action instead passing functions back and forth.

What do you think?

app/js/views/themes-list-item.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Delete a recipe.
+   */
+  handleOnDelete() {
+    this.foxbox.recipes.remove(this.props.theme.id)
+      .then(() => {
+        this.props.update();
+      })
+      .catch(console.error.bind(console));
```
azasypkin: __nit:__ subjectively ```.catch((e) => console.error(e))``` feels more friendly than using ```bind```, but up to you.

app/js/views/themes-new.jsx
```diff
@@ -2,235 +2,227 @@ import React from 'components/react';
...
-        id: 3,
-        name: 'motion sensor'
-      },
-      {
-        id: 4,
-        name: 'front door'
-      }
-    ];
-    this.setState({ services });
+  updateServices(services = [[], []]) {
```
azasypkin: __nit:__ wondering if it's possible to do like ```updateServices([servicesWithGet, servicesWithSet] = [[], []])```

app/js/views/themes-new.jsx
```diff
@@ -2,235 +2,227 @@ import React from 'components/react';
...
+    this.foxbox.recipes.add({
+        label,
+        serviceWithGet: serviceWithGetObject.id,
+        getter: getterObject.value,
+        serviceWithSet: serviceWithSetObject.id,
+        setter: setterObject.value,
         enabled: true
       })
       .then(() => {
+        console.log(arguments);
```
azasypkin: __nit:__ leftover

app/js/views/themes-new.jsx
```diff
@@ -2,235 +2,227 @@ import React from 'components/react';
...
-    if (this.state.property !== null) {
+    if (this.state.getter !== null) {
       className += ' new-theme__select--selected';
     }
     let propertySelector = (
       <select className="new-theme__select new-theme__select--hidden"></select>
     );
-    if (this.state.service !== null) {
+    if (this.state.serviceWithGet !== null) {
+      optionNodes = this.state.servicesWithGet[this.state.serviceWithGet].get
```
azasypkin: When I select Ip camera, it complains with ```TypeError: this.state.servicesWithGet[this.state.serviceWithGet].get.map is not a function```

azasypkin:
> @azasypkin: This patch is a bit messy with a lot of code duplication and saving a rule to the box doesn't seem to work.
But if you agree, let's get it landed first and we can fix and refactor it. It's already big.

I've left few questions and most obvious nits. You can handle as much as you like. Since your changes are mostly isolated I'm fine if you land it, so we can improve it continuously, step-by-step :)

app/css/views/themes.css
gmarty: I added the x because of a flash of unloaded image, but the `font-size` should be sufficient.

app/js/lib/foxbox/recipes.js
gmarty: I want to integrate it directly in `this[p.net].fetchJSON()` so that API calls are less verbose.

app/js/lib/foxbox/recipes.js
gmarty: I think I prefer the second case actually. Putting on several lines make it better for readibility.

app/js/lib/foxbox/recipes.js
gmarty: That's a remaining of when there was a `console.log` here!

app/js/lib/foxbox/recipes.js
gmarty: I didn't know about this API. Let me update the code.

app/js/lib/foxbox/recipes.js
gmarty: I know but the REST API is not user-friendly enough to be expose to the client directly.
Ideally we want the foxbox lib to output user-friendly data and then localise in the view as you suggest.

app/js/lib/foxbox/recipes.js
gmarty: I thought `getters` and `setters` were too verbose.

app/js/lib/foxbox/recipes.js
gmarty: Yep, as soon as this works!

app/js/lib/foxbox/recipes.js
gmarty: Ah yeah good idea!

app/js/lib/foxbox/recipes.js
gmarty: Nope, not as today.

app/js/lib/foxbox/recipes.js
azasypkin: Mmm, I'm confused :) ```Putting on several lines``` is the first, not second... Or what do you mean?

app/js/views/themes.jsx
gmarty: Yes I agree. Theme is only used in the route and in the names of the views.

app/js/views/themes.jsx
gmarty: Surprisingly passing a cb as a prop is a common pattern in React, see https://facebook.github.io/react/tips/communicate-between-components.html
They recommend an event system when the components don't share a parent-child relationship.

I'm leaving it for now, and we'll change the code when we adopt an event approach.

app/js/views/themes-list-item.jsx
gmarty: I like the `bind` pattern in general because it will display all the arguments if there are more than one.

app/css/views/themes.css
azasypkin: > And yes, we need to work a little on accessibility, but I don't have enough knowledge in that field.

Ah, np, I think we can plan accessibility week/days if we get approval for the next foxbox stage :)

app/js/lib/foxbox/recipes.js
azasypkin: Yeah, I still feel we'll have use cases when we need to assign url to the ```img[src]``` or ```a[download]``` - so let's keep this in mind.

app/js/views/themes-new.jsx
gmarty: I'll enquire it when I have time to set up the IP Camera here.

app/js/lib/foxbox/recipes.js
azasypkin: Ugh, it's a shame :/

app/js/views/themes-list-item.jsx
azasypkin: In the approach without bind you also see all the arguments, but the problem in this particular case is that it's not obvious that there is ```e``` argument :) But it's not critical to me anyway, just wanted to note

app/js/lib/foxbox/recipes.js
gmarty: Actually that's a significant refactoring that will impact the view too.
I'll do it later.

app/js/views/themes-new.jsx
gmarty: It looks like it's working!

app/js/views/themes.jsx
azasypkin: Grrrr, I see, so they left it up to developers :/

PR #52, opened 2016-04-05
Deactivate transpilation to es5. r=azasypkin

gmarty:
We're targeting the latest versions of Fennec and Chrome Android for now, so we can safely deactivate the transpilation to es5.
Babel is used for es2015 modules and JSX conversion. In the future, we'll activate missing es2015 features as we support a wider range of browsers.

@azasypkin: do you approve the above?


azasypkin:
r+, let's see how it goes!

PR #53, opened 2016-04-05
Implement the new design colours and icons. r=ferjm

gmarty:


app/js/views/navigation-menu.jsx
```diff
@@ -20,19 +20,67 @@ export default class NavigationMenu extends React.Component {
...
-      ['themes', 'Themes'],
-      ['mr-fox', 'Mr. Fox']
+      {
+        id: 'services',
+        label: 'Home',
+        icon: (
+          <svg xmlns="http://www.w3.org/2000/svg" height="30" width="30">
+            <path d="M3 27h8.256v-9.39h7.488V27H27V11.435L15 3 3 11.435V27z"
+                  fill="#fff"/>
+          </svg>
```
gmarty: @azasypkin: I embedded the SVG files into the JSX to be able to style the `path`s using CSS (the icons are white by default and blue when active). I'd like your opinion on this approach. I don't really like it but I can't find of any other way to style icon in CSS.

gmarty: Or maybe I should just use CSS filter...

azasypkin: Yeah, or maybe we can leverage blend mode or something to simulate active state with original image? Some day in SMS we agreed with VD to use something like http://azasypkin.github.io/style-my-image/ to change image states with CSS only without second image, wondering if we can do the same now.

ferjm:
LGTM. There are some test issues though

PR #54, opened 2016-04-05
Introduce Base view. r=gmarty

azasypkin:


azasypkin:
Hey @gmarty may I have your early feedback before moving forward/changing direction or even abandoning the idea? :)

The idea is the following (somewhat similar to ASP.NET master page):
* We have a master view component, that defines elements that are common across the app;
* Master view defines slots/placeholders where particular app view can render something. In our case it's only ```body``` part;
* Controller creates Master view component and configures the body (defines which Component will be rendered into ```body``` slot and which properties it needs);
* In addition to view-specific properties that controller provides for the ```body``` component, master view always adds properties that are needed by every view, eg. ```foxbox``` instance;
* In addition to that app view would like to modify header/footer (hide, style or even add buttons/commands). For that purpose master view provides body component with the following methods: ```configureHeader(object)``` and ```configureFooter(object)``` (ideally it should be handled by event model somehow, but as we figured out there is no recommended/built-in way to do that);

In the current PR I've migrated only login/services/service views (tests are red) and didn't touch recipes views as they will need custom buttons in the header. So I'd be happy to get your vision/feedback before introducing more complexity :)

app/js/views/master-view.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+const DEFAULT_HEADER_CONFIG = Object.freeze({
+  visible: true
+});
+
+const DEFAULT_FOOTER_CONFIG = Object.freeze({
+  visible: true
+});
+
+export default class MasterView extends React.Component {
```
gmarty: Usually this concept is referred to as layout. So maybe rename it to `LayoutView` or just `Layout`.

app/js/views/services.jsx
```diff
@@ -1,8 +1,6 @@
 import React from 'components/react';
 
-import NavigationMenu from 'js/views/navigation-menu';
 import ServicesList from 'js/views/services-list';
```
gmarty: nit: So do we want to append `View` to all the views class names? If so, let's rename it to `ServicesListView`.

gmarty:
I like the approach globally.
My only concern is that it's going to be too complicated or too restrictive for the header. I can see 2 solutions:
* Pass React nodes to the header instead of parameters. The nodes should be defined in the main component view class (and can possibly be reset at a sub view level).
* Simply make a dummy Header view class that each main component will use in their body. This is what's going to give us the more simplicity and flexibility. The dummy Header view will just wrap the content into a `<header className={className}>{headerNode}</header>`. The UX specs requires a complex logic in the header connected to the rest of the page, so I prefer this approach.

The logic you implemented for the footer works well because it is either hidden or visible. No need for any other type of logic.

I like how the top component views are finally getting slimmer without all the layout clutter!

app/js/views/master-view.jsx
azasypkin: It's going to do more than just layout (manage modals for example), but I don't really care about naming - let's make it ```LayoutView``` then :)

app/js/views/services.jsx
azasypkin: I'd vote for it, yes, IMO it's much clearer. Will rename

azasypkin:
> My only concern is that it's going to be too complicated or too restrictive for the header. I can see 2 solutions:
The logic you implemented for the footer works well because it is either hidden or visible. No need for any other type of logic.

I had something in mind for this already, let me see what works best for us.

azasypkin:
Well, I gave it some thought during PTO and have to admit that it maybe too early to generalize too much as we haven't implemented enough different cases yet to at least see consistent pattern. So I've chosen another approach that it's optimized for the most frequent cases and still allows us to fully customize both header and footer on per view basis. We can consider this as the first step forward to great&shiny generalization :)

r? @gmarty

app/js/views/service.jsx
```diff
@@ -19,18 +20,26 @@ export default class Service extends React.Component {
     });
   }
 
-  render() {
+  renderHeader() {
+    return super.renderHeader(
+      this.state.service && this.state.service.properties.name ?
+        this.state.service.properties.name :
+        'Unknown Service'
```
gmarty: What about:
```js
this.state.service.properties.name || 'Unknown Service'
```

app/js/views/services.jsx
```diff
@@ -65,25 +63,15 @@ export default class Services extends React.Component {
...
-        <section className="app-view__body">
-          <h2>General</h2>
-          <ServicesList services={this.state.services} foxbox={this.foxbox}/>
-        </section>
-        <footer className="app-view__footer">
-          <NavigationMenu foxbox={this.foxbox}/>
-        </footer>
-        <Modal visible={this.state.isModalVisible} title={this.state.title}
-               body={this.state.body} dismiss={this.dismissModal.bind(this)}/>
+      <div className="app-view__body-placeholder">
```
gmarty: Do we really need the `app-view__body-placeholder` class here? It looks like it's going to be something easy to forget if we want to add more top views.

app/js/views/service.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'components/react';
 
-import CameraService from 'js/views/services/camera';
-import DefaultService from 'js/views/services/default';
+import BaseView from 'js/views/base-view';
+import CameraServiceView from 'js/views/services/camera';
+import DefaultServiceView from 'js/views/services/default';
 
-export default class Service extends React.Component {
+export default class Service extends BaseView {
```
gmarty: Do we want to rename the top view classes too? i.e. `ServiceView`  here.
Maybe in the future we want to store them in a dedicated folder (like `views/top/`).

gmarty:
First of all, I really like this approach. I think it just makes sense and it's clean.

To avoid confusion with other types of views, I'd give them particular names like (`ServiceTopView`/`service-top.jsx` or similar) and maybe isolate them in their own folder.
I also would like to avoid repetitions of the `app-view__body-placeholder` class name in the body wrapper, but I'm fine with that if it's better for the BEM logic and CSS organisation.

Finally, I'd like to see this patch landing sooner rather than later and see how it works for us. We can always iterate if it needs some tweaking. Is it ready for production?
r+

app/js/views/service.jsx
azasypkin: But we had in ```if(!this.state.service)``` in ```renderBody``` I didn't check but I thought that there is a chance that render is called when we don't have service instance at all?

app/js/views/services.jsx
azasypkin: I don't like it as much as you do :) But here is my rationale:

I want main class (```app-view__body```) to be defined in base view as it's base view responsibility to define view structure including "main" css classes, but then every view should return single element in ```renderBody``` (because of [1]), so ```app-view__body-placeholder``` is not a requirement, but rather an optional thing - you use it only if you want your wrapper node to fill entire ```app-view__body``` section. I can think of two alternatives:
* Rename ```app-view__body-placeholder``` to something more obvious like ```app-view_fill-body``` or something to stress that it's styling only and doesn't have any functional meaning;
* Remove this class and if view wants its body to fill entire ```app-view__body``` it defines its own styling;
* Or we can force it with something ```app-view__body > :first-child``` - but I doooon't like this option :) It's not BEM'y, not performant and make a fixed decision on behalf of the particular view itself.

What do you think @gmarty ?

[1] http://facebook.github.io/react/tips/maximum-number-of-jsx-root-nodes.html

app/js/views/service.jsx
azasypkin: > Do we want to rename the top view classes too? i.e. ServiceView here.

I'd do it, but ideally in the consequent patches, like ```changing view - change it name``` :) And as you said we'll have time to think how we want to organize views.

azasypkin:
> To avoid confusion with other types of views, I'd give them particular names like (ServiceTopView/service-top.jsx or similar) and maybe isolate them in their own folder.

Yep, we can do something like this, I think after this rush demo we can "sit together" and work out good view structure :)

> I also would like to avoid repetitions of the app-view__body-placeholder class name in the body wrapper, but I'm fine with that if it's better for the BEM logic and CSS organisation.

Yeah, I agree, I replied with rationale and options, please tell me what you prefer.

> Is it ready for production?

It's! Let me know your thoughts and I'll land it based on your replies :)

app/js/views/services.jsx
gmarty: Thanks for the detailed explanation. I have a preference for the first solution, but I leave it to you to decide whether to change something or not.

gmarty:
OK let's land this and we'll take the time later to discuss the organisation of view files.

app/js/views/services.jsx
azasypkin: > I have a preference for the first solution

Deal!

azasypkin:
Renamed ```app-view_body-placeholder``` to ```app-view_fill-body``` and travis is green.

Thanks for review!

PR #69, opened 2016-04-12
Minor changes in the recipes

gmarty:
This patch contains:
* User friendly name for the IP camera
* Filter out the console adapter

### DIFF
--- a/app/js/lib/foxbox/recipes.js
+++ b/app/js/lib/foxbox/recipes.js
@@ -28,8 +28,9 @@ export default class Recipe {
       `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
       // Get all recipes.
       .then(services => services.filter(service =>
-        service.adapter === 'thinkerbell-adapter' &&
-        service.id !== 'thinkerbell-root-service'
+        service.adapter !== 'thinkerbell-adapter'
+          && service.adapter !== 'webpush@link.mozilla.org'
+          && service.adapter !== 'console@link.mozilla.org'
       ))
       // Fetch their respective enabled status.
       .then(recipes => {
@@ -52,7 +53,8 @@ export default class Recipe {
             id: recipe.id,
             label: recipe.id,
             enabled: servicesEnabled[index][`${recipe.id}/get_enabled`]
-              .OnOff === 'On'
+              && servicesEnabled[index][`${recipe.id}/get_enabled`]
+                .OnOff === 'On'
           })));
       });
   }
@@ -62,8 +64,12 @@ export default class Recipe {
       `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
       // Get all services but the recipes.
       .then(services => services.filter(service =>
-        service.adapter != 'thinkerbell-adapter' &&
-        service.adapter != 'webpush@link.mozilla.org'
+        service.adapter !== 'thinkerbell-adapter'
+          && service.adapter !== 'webpush@link.mozilla.org'
+          && service.adapter !== 'console@link.mozilla.org'
+
+          // Deactivating IP camera from services with getters.
+          && service.adapter !== 'ip-camera@link.mozilla.org'
       ))
       // Keep only the services with getters.
       .then(services => services.filter(service =>
@@ -71,13 +77,13 @@ export default class Recipe {
       ))
       // User friendly name.
       .then(services => services.map(service => {
-          switch (service.id) {
-            case 'service:clock@link.mozilla.org':
+          switch (service.adapter) {
+            case 'clock@link.mozilla.org':
               service.name = 'Everyday';
               break;
 
             default:
-              service.name = service.id;
+              service.name = service.adapter;
               break;
           }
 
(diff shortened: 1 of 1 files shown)


app/js/lib/foxbox/recipes.js
```diff
@@ -28,8 +28,9 @@ export default class Recipe {
       `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
       // Get all recipes.
       .then(services => services.filter(service =>
-        service.adapter === 'thinkerbell-adapter' &&
-        service.id !== 'thinkerbell-root-service'
+        service.adapter != 'thinkerbell-adapter'
+        && service.adapter != 'webpush@link.mozilla.org'
```
azasypkin: __nit:__ when boolean condition is wrapped it's usually indented to keep to make it easier to track single condition, like:
```
service.adapter != 'thinkerbell-adapter'
  && service.adapter != 'webpush@link.mozilla.org'
  && service.adapter != 'console@link.mozilla.org'
```

And we need to agree on ```!==``` or ```!=```, triple sign feels safer in JS world, but I'm fine with any choice :)

app/js/lib/foxbox/recipes.js
```diff
@@ -62,22 +63,30 @@ export default class Recipe {
...
       // Get all services but the recipes.
       .then(services => services.filter(service =>
-        service.adapter != 'thinkerbell-adapter' &&
-        service.adapter != 'webpush@link.mozilla.org'
+        service.adapter != 'thinkerbell-adapter'
+        && service.adapter != 'webpush@link.mozilla.org'
+        && service.adapter != 'console@link.mozilla.org'
+
+          // Deactivating IP camera from services with getters
+        && service.adapter != 'ip-camera@link.mozilla.org'
```
azasypkin: __nit:__ same indent note here, + comment is misaligned + comments should end with full stop :)

app/js/lib/foxbox/recipes.js
```diff
@@ -62,22 +63,30 @@ export default class Recipe {
...
       // User friendly name.
       .then(services => services.map(service => {
-          switch (service.id) {
-            case 'service:clock@link.mozilla.org':
+          switch (service.adapter) {
+            case 'clock@link.mozilla.org':
               service.name = 'Everyday';
               break;
 
+            case 'ip-camera@link.mozilla.org':
```
azasypkin: __question:__ mm I'm a bit puzzled, it looks like we've just filtered out ip camera service above?

app/js/lib/foxbox/recipes.js
gmarty: I usually ban `==` and always prefer `===` (or `!==`). I didn't notice that ones. Let's enforce that with eslint.

I placed the operator at the beginning of the new line to avoid polluting the git diff when we add new conditions (only one line change), but I find it weird. Tell me what you prefer and I'll use it.

app/js/lib/foxbox/recipes.js
```diff
@@ -128,22 +137,27 @@ export default class Recipe {
       `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
       // Get all services but the recipes.
       .then(services => services.filter(service =>
-        service.adapter != 'thinkerbell-adapter' &&
-        service.adapter != 'webpush@link.mozilla.org'
+        service.adapter != 'thinkerbell-adapter'
```
azasypkin: __nit:__ same indent note here

app/js/lib/foxbox/recipes.js
gmarty: I always use the style formatting on WebStorms. I fine tuned it to fit my needs, but it's not perfect and in some case it results in weird indentations like here. I'll fix that.

app/js/lib/foxbox/recipes.js
gmarty: ah yeah correct. This can go away.

app/js/lib/foxbox/recipes.js
```diff
@@ -173,6 +187,16 @@ export default class Recipe {
               ];
               break;
 
+            case 'ip-camera@link.mozilla.org':
+              service.setters = [
+                {
+                  label: 'takes a picture',
+                  value: 'setter:snapshot.ae67e622-7a66-465e-bab0-' +
```
azasypkin: ugh, hardcoding such things is bad, but let's move forward with this and will change after this big demo :/ Anyway I will be poking around this piece of code right now.

azasypkin:
r+ with green travis, nits handled, question answered and issue fixed (I see ```TypeError: servicesEnabled[index][((intermediate value) + "/get_enabled")] is undefined``` when I try to save recipe, that I didn't see before) :)

app/js/lib/foxbox/recipes.js
azasypkin: Honestly, I'd prefer this one, it's easier to read and see that all statements belong to the same condition, regarding git diff - if we add new conditions we just add more lines and previous lines should not be affected 
```
service.adapter != 'thinkerbell-adapter'
  && service.adapter != 'webpush@link.mozilla.org'
  && service.adapter != 'console@link.mozilla.org'
```

gmarty:
I just fixed this error and now I can see the recipes.

Issue #47, opened 2016-04-01
Use eslint to enforce coding styles


gmarty:
[comma-dangle](http://eslint.org/docs/rules/comma-dangle#always-multiline)
[arrow-parens](http://eslint.org/docs/rules/arrow-parens#always)

The rationale behind the coding styles is that it should result in cleaner git diff and code easier to read.

azasypkin:
Let's add ```guard-for-in``` rule as well!

PR #70, opened 2016-04-12
Add support for the Philips Hue adapter. r=azasypkin

gmarty:
This patch:
* Add partial support for the Philips Hue adapter (getting values is not done yet)
* Relax the use of eslint `comma-dangle` in preparation of https://github.com/fxbox/app/issues/47
* Fix minor issue in `foxbox.js` (better warning when no boxes are found and only poll the box when the user is logged in)
* Remove the code related to the deprecated service states

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,7 +24,8 @@
     "no-use-before-define": ["error", { "functions": false, "classes": true }],
     "no-var": ["error"],
     "quotes": ["error", "single"],
-    "semi": ["error", "always"]
+    "semi": ["error", "always"],
+    "comma-dangle": ["warn", "only-multiline"]
   },
 
   "env": {
--- a/app/js/lib/foxbox/foxbox.js
+++ b/app/js/lib/foxbox/foxbox.js
@@ -156,6 +156,13 @@ export default class Foxbox extends Service {
    * @param {number} index The index of the box in the boxes array.
    */
   selectBox(index = 0) {
+    if (!this[p.boxes].length) {
+      this[p.settings].configured = false;
+      console.error('No boxes found. Is this app online? Is the box online?');
+
+      return;
+    }
+
     if (index >= this[p.boxes].length) {
       this[p.settings].configured = false;
       console.error('Index out of range.');
@@ -254,8 +261,9 @@ export default class Foxbox extends Service {
    */
   schedulePoll() {
     // Return early if polling is not enabled or it has already been scheduled.
-    if (!this[p.isPollingEnabled] ||
-      this[p.nextPollTimeout]) {
+    if (!this[p.isPollingEnabled]
+      || this[p.nextPollTimeout]
+      || !this.isLoggedIn) {
       return;
     }
 
@@ -350,44 +358,6 @@ export default class Foxbox extends Service {
       });
   }
 
-  /**
-   * Fetch the state of a service from the box.
-   *
-   * @param {string} id The ID of the service.
-   * @return {Promise}
-   */
-  getServiceState(id) {
-    return this[p.net].fetchJSON(`${this[p.net].origin}/services/${id}/state`)
-      .then(res => {
-        if (!res) {
-          throw new Error('The action couldn\'t be performed.');
-        }
-
-        return res;
-      });
-  }
-
-  /**
-   * Change the state of a service.
-   *
-   * @param {string} id The ID of the service.
-   * @param {Object} state An object containing pairs of key/value.
-   * @return {Promise}
-   */
-  setServiceState(id, state) {
-    return new Promise((resolve, reject) => {
-      this[p.net].fetchJSON(`${this[p.net].origin}/services/${id}/state`,
-        'PUT', state)
-        .then(res => {
-          if (!res || !res.result || res.result !== 'success') {
-            return reject(new Error('The action couldn\'t be performed.'));
-          }
-
-          return resolve();
-        });
-    });
-  }
-
   getTags() {
     return this[p.db].getTags.apply(this[p.db], arguments);
   }
@@ -473,6 +443,8 @@ export default class Foxbox extends Service {
     switch (operationKind) {
       case 'TakeSnapshot':
         return 'Unit';
+      case 'LightOn':
+        return 'OnOff';
       default:
         return operationKind;
     }
--- a/app/js/lib/foxbox/settings.js
+++ b/app/js/lib/foxbox/settings.js
@@ -28,7 +28,8 @@ const QUERY_STRING_AUTH_TOKEN_NAME = 'auth';
 const storage = localStorage ? localStorage : {
   getItem: () => undefined,
   setItem: () => {},
-  removeItem: () => {}
+  removeItem: () => {},
+  clear: () => {},
 };
 
 export default class Settings extends Model {
@@ -62,6 +63,8 @@ export default class Settings extends Model {
     return new Promise(resolve => {
       // @todo Remove only the items set here.
       storage.clear();
+      this._session = null;
+
       resolve();
     });
   }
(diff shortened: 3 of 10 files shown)


.eslintrc
```diff
@@ -24,7 +24,8 @@
     "no-use-before-define": ["error", { "functions": false, "classes": true }],
     "no-var": ["error"],
     "quotes": ["error", "single"],
-    "semi": ["error", "always"]
+    "semi": ["error", "always"],
+    "comma-dangle": ["warn", "only-multiline"]
```
azasypkin: __nit:__ Maybe let's enforce rule with ```"error"```?

Also I thought your ultimate goal was something more like ```["error", "always-multiline"]``` or I just misunderstood?

gmarty: No that's correct. But for now I just want to allow it. We'll enforce this rule later when working on #47.

app/js/lib/foxbox/foxbox.js
```diff
@@ -260,6 +267,10 @@ export default class Foxbox extends Service {
     }
 
     this[p.nextPollTimeout] = setTimeout(() => {
+      if (!this.isLoggedIn) {
```
azasypkin: __nit:__ maybe just add this new condition to the ```if``` branch above (```if (!this[p.isPollingEnabled] || this[p.nextPollTimeout] || !this.isLoggedIn) ```)? In this case we'll have it inside two main places at the very beginning (```schedulePoll``` and ```refreshServicesByPolling```)

app/js/lib/foxbox/foxbox.js
```diff
@@ -350,44 +361,6 @@ export default class Foxbox extends Service {
       });
   }
 
-  /**
-   * Fetch the state of a service from the box.
-   *
-   * @param {string} id The ID of the service.
-   * @return {Promise}
-   */
-  getServiceState(id) {
```
azasypkin: \0/ Then maybe remove related dead code from ```refreshServicesByPolling```?

app/js/lib/foxbox/foxbox.js
```diff
@@ -473,6 +446,8 @@ export default class Foxbox extends Service {
     switch (operationKind) {
       case 'TakeSnapshot':
         return 'Unit';
+      case 'power':
```
azasypkin: __note:__ Just wanted to note that lower case ```power``` looks suspicious :)

app/js/lib/foxbox/settings.js
```diff
@@ -62,6 +62,8 @@ export default class Settings extends Model {
     return new Promise(resolve => {
       // @todo Remove only the items set here.
       storage.clear();
```
azasypkin: __note:__ also not related to your patch, but it's funny that fake storage defined above doesn't have ```clear``` method, so with fake storage this promise will never resolve :)

app/js/views/service.jsx
```diff
@@ -33,13 +34,16 @@ export default class Service extends BaseView {
...
         return (<CameraServiceView service={this.state.service}
-                                   foxbox={this.foxbox} />);
+                                   foxbox={this.foxbox}/>);
+      case 'philips_hue@link.mozilla.org':
+        return (<LightServiceView service={this.state.service}
+                                  foxbox={this.foxbox}/>);
       default:
         return (<DefaultServiceView service={this.state.service}
-                                    foxbox={this.foxbox} />);
+                                    foxbox={this.foxbox}/>);
```
azasypkin: __question:__ why do you remove space before ```/>```? I was usually adding them to separate it from the value of the last attribute (haven't checked w3c spec on this though, but browsers are not restrictive at all). Let's agree on the same approach! What do you prefer? :)

app/js/views/services-list-item.jsx
```diff
@@ -5,24 +5,57 @@ export default class ServicesListItem extends React.Component {
...
+    const id = this.getOperationByAlias('power').id;
+
+    // Optimistic update.
+    this.setState({ on });
+
+    this.foxbox.performSetOperation({
+        id,
+        kind: 'power'
+      }, value)
+      .catch((error) => {
```
azasypkin: __nit:__ alignment is somewhat messy, parameter object is on the same level as ```.catch``` branch

app/js/views/services-list-item.jsx
```diff
@@ -5,24 +5,57 @@ export default class ServicesListItem extends React.Component {
...
+        // Revert back to the previous value.
+        this.setState({ on: !on });
+        console.error(error);
```
azasypkin: __nit:__ would be nice to have more detailed error description (errors returned from taxonomy are a bit vague), like:
```
console.error('Error occurred while blah-blah light switch blah-blah', error);
```

app/js/views/services-list-item.jsx
```diff
@@ -5,24 +5,57 @@ export default class ServicesListItem extends React.Component {
...
   handleLightOnChange(evt) {
-    // No optimistic update because some network issues can't be caught.
     const on = evt.target.checked;
+    const value = on ? 'On' : 'Off';
+    const id = this.getOperationByAlias('power').id;
+
+    // Optimistic update.
+    this.setState({ on });
+
+    this.foxbox.performSetOperation({
```
azasypkin: __question:__ mmm I don't get why can't you just directly pass here operation returned from ```getOperationByAlias```?

app/js/views/services/light.jsx
```diff
@@ -0,0 +1,96 @@
...
+    name = name.trim();
+    this.foxbox.setTag({ name })
+      .then(() => {
+        this.populateTags(); // Needed to get the newly added tag ID.
+      });
+  }
+
+  render() {
+    return (
+      <div className="app-view">
```
azasypkin: __issue:__ Ah, sorry I'll need to account for new BaseView, so your render should just render body (probably with ```app-view__fill-body``` class)

app/js/views/services/light.jsx
```diff
@@ -0,0 +1,96 @@
...
+  onServiceStateChanged(service) {
+    if (service.id !== this.service.id) {
+      return;
+    }
+
+    this.service = service;
+    this.setState({ data: service.tags });
+  }
+
+  populateTags() {
```
azasypkin: Mm, we'll we have tag management functionality in Light service page?

azasypkin:
Okay looks good to me! I don't have Hue to check the lights, so I was just looking at the code part + that default and camera services are working properly (with yesterdays foxbox).

__r+__ with nits handled, questions answered, Light service view fixed (it should be migrated to BaseView approach) and green Travis!

Thanks

app/js/lib/foxbox/foxbox.js
gmarty: I'll leave the code related to polling for now, but I'm planning on refactoring it for https://github.com/fxbox/app/pull/73.

app/js/views/service.jsx
gmarty: I think I must have configured my IDE in the past to remove the whitespace before `/>`.
It doesn't change anything once the code is compiled from JSX to plain JS though.

app/js/lib/foxbox/foxbox.js
gmarty: Ah it's been renamed `LightOn` in the latest version.

app/js/views/services-list-item.jsx
gmarty: I haven't realised I could do that.

app/js/lib/foxbox/settings.js
gmarty: True. I'm going to fix it in this PR or I'll just forget about it!

app/js/views/services/light.jsx
gmarty: It's the fake tags for now, purely frontend. I'll keep them for now but they'll probably go away soon.

PR #73, opened 2016-04-12
Add specific classes for each service type. r=azasypkin

gmarty:


gmarty:
With this patch, the tag system (in their current front-end implementation, that is IndexedDB backed) does not work anymore.

app/js/lib/foxbox/services/light.js
```diff
@@ -0,0 +1,117 @@
...
+
+    return this.performSetOperation(operation, value);
+  }
+
+  /**
+   * Turn on the bulb.
+   *
+   * @return {Promise}
+   */
+  turnOn() {
```
azasypkin: Just read the article from microsoft and they call our service wrappers - "device translators" [1] :)

[1] https://github.com/openT2T/translators/blob/master/org.OpenT2T.Sample.SuperPopular.Lamp/Philips%20Hue%20Light%20Bulb/js/thingTranslator.js#L76-L106

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,147 @@
+'use strict';
+
+const TYPE = 'Unknown';
+
+// Private members.
```
azasypkin: __nit:__ It's nitpicking, but seems ```// Private members.``` doesn't bring any value here :) With the following comments it's super clear what this is for :)

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,147 @@
...
+   * @private
+   */
+  [p.getOperationValueType](operationKind) {
+    if (!operationKind) {
+      throw new Error('Operation kind is not defined.');
+    }
+
+    // Operation kind can be either object or string.
+    if (typeof operationKind === 'object') {
+      return operationKind.typ;
```
azasypkin: __issue:__ there is no ```typ``` anymore

app/js/lib/foxbox/services/light.js
gmarty: Oh the API is similar!
I don't really like device translator. Specifically because we use the term `service` and `translator` doesn't bring any meaning here.
I like `service wrapper` though. Maybe we should use that term to refer to these classes.

azasypkin: I was just kidding and wanted to share with you :) I mean I don't really like how their code looks like right now (especially this "translator" term), I prefer our approach :)

app/js/lib/foxbox/services/ip-camera.js
```diff
@@ -0,0 +1,86 @@
...
+  getImageList() {
+    return this[p.query]('image_list');
+  }
+
+  getLatestImage() {
+    return this[p.query]('latest image');
+  }
+
+  // Public methods.
+  takeSnapshot() {
```
azasypkin: Let me describe what I have in my head so that we can discuss and workout best possible solution :)

I was thinking about the approach when the most of heavy lifting is done in base service, while specific service know just a few things. Let's take a camera service as example. To take a snapshot we have the following method:
```
const operation = this.getOperation(this[p.setters], 'TakeSnapshot');

return this.performSetOperation(operation)
  .then(() => {
    return this.performGetOperation(
      this.getOperation(this[p.getters], 'latest image')
    );
  });
```

Looking at this method I feel unnecessary step - we retrieve operation from base service and then immediately pass it to base service again. That means we can improve it, like this:

```
this.performSetOperation({ kind: 'TakeSnapshot'})
  .then(
    () => this.performGetOperation({ kind: 'latest image' })
  );
```

Basically specific service don't need to have ability to retrieve operation itself, it's too much technical details and too coupled to the current API state. I use ```{ kind: 'TakeSnapshot' }``` cause I was thinking there maybe cases when we want to query channel (I'm trying to use vocabulary from taxonomy where getters and setters are channels) using channel id, in this case we'll have ```{ id: 'clock@adapter' }```.

As in any API we can leverage polymorphous methods to have sensible defaults. Let's say channel kind is the most frequently used channel selector so we can simplify even further:
```
this.performSetOperation('TakeSnapshot')
  .then(() => this.performGetOperation('latest image'));

// Or if we use setter-getter-vocabulary.
// Too confusing though, maybe send/receive or just use current.
this.set('TakeSnapshot').then(() => this.get('latest image'));
``` 
To support this base service should have something like this (semi-pseudo code):
```
set(selector, value) {
  const setter = this[p.getChannel](this[p.service].setters, selector);

  const operationType = this[p.getChannelValueType](setter.kind);

  return this[p.net].fetchJSON(
    `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/set`,
    'PUT',
    [[{ id: setter.id }, { [operationType]: value }]]
  );
}

[p.getChannel](channels, selector) {
  let channelKey;

  if (selector.id) {
    channelKey = selector.id;
  } else if (selector.kind || typeof selector === 'string') {
    // That is optional default I was talking above.
    const channelKind = selector.kind || selector;
    channelKey = Object.keys(channels).find((key) => {
      const channel = channels[key];

      if (typeof channel.kind === 'object') {
        return channel.kind.kind === channelKind;
      }

      return channel.kind === channelKind;
    });
  }

  return channelKey ? channels[channelKey] : null;
}
```

Also theoretically we can pass ```selector``` directly to the service as getter/setter selector - but it's dependency on the API, so I'm really sure.

It's just an idea I had in mind since our last f2f discussion. What do you think? Am I missing something?

gmarty: I really like the idea of moving as much code as possible to the base class. I also think simplifying the code makes a lot of sense.

I'm not so much attached to the internals than to the publicly exposed API. If doing these changes means less effort when writing classes for new adapters, I'm supporting it!

Do you want me to implement it before you continue the review?

app/js/lib/foxbox/services/ip-camera.js
```diff
@@ -0,0 +1,86 @@
...
+    return this[p.model];
+  }
+
+  get name() {
+    return this[p.name];
+  }
+
+  // Public dynamic property getters.
+  getImageList() {
+    return this[p.query]('image_list');
```
azasypkin: __issue:__ looks like there is no ```query``` symbol defined?

app/js/lib/foxbox/services/ip-camera.js
```diff
@@ -0,0 +1,86 @@
...
+  constructor(props, config) {
+    super(props, config);
+
+    // Public getter only properties.
+    this[p.manufacturer] = props.properties.manufacturer;
+    this[p.model] = props.properties.model;
+    this[p.name] = props.properties.name;
+
+    // Private getter only properties.
+    this[p.adapter] = props.adapter;
```
azasypkin: __question:__ what is the benefit of having separate properties vs eg. ```this[p.service]``` (and then ```this[p.service].getters``` or ```this[p.service].id```) ?

gmarty: I thought it's easier to read and we only keep the properties we need (better code hygiene) but I don't have strong opinion.
If you strongly disagree, I'll adopt the second approach like `this[p.service].getters`.

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,156 @@
...
+    this[p.net] = config.net;
+    this[p.settings] = config.settings;
+
+    // Public getter only properties.
+    this[p.id] = props.id;
+    this[p.getters] = props.getters;
+    this[p.setters] = props.setters;
+
+    // Public properties.
+    this._tags = props.tags;
```
azasypkin: __nit:__ looks like nobody use it and underscore based name is weird

app/js/lib/foxbox/services/ip-camera.js
azasypkin: No, I don't have strong opinion either, let's keep it as is for now.

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,156 @@
+'use strict';
+
+const TYPE = 'Unknown';
```
azasypkin: __optional nit:__ If it's not user facing string, maybe it will be more convenient to use lower-case only strings (```unknown, ip-camera, etc```)? Maybe it's not a big deal in the end, but sometimes it costs developer time to notice case-related typo :)

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,156 @@
...
+  get id() {
+    return this[p.id];
+  }
+
+  get name() {
+    return '';
+  }
+
+  get hasGetters() {
+    return Object.keys(this[p.getters]).length > 0;
```
azasypkin: __note:__ I have mixed feelings about potentially heavy method calls inside getter (as it masks operations from the consumer, usually getter implies nearly 0-cost operation, although browser API does it all the time :/), so I usually use real methods/functions instead. But in our case it won't likely be a problem, so up to you.

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,156 @@
...
+
+  get hasSetters() {
+    return Object.keys(this[p.setters]).length > 0;
+  }
+
+  // Public methods.
+  /**
+   * Call a service setter with a value.
+   *
+   * @param {Object} selector
```
azasypkin: __nit:__ Maybe just a simple jsdoc for parameters?

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,156 @@
...
+   */
+  get(selector) {
+    const getter = this[p.getChannel](this[p.getters], selector);
+    const body = { id: getter.id };
+
+    if (getter.kind.type === 'Binary') {
+      return this[p.net].fetchBlob(
+        `${this[p.net].origin}/api/v${this[p.settings].apiVersion}` +
+        '/channels/get',
+        // For now we only support JPEG blobs.
```
azasypkin: __issue__: hmm, looks inappropriate for the base class getter... Does your lights getter work correctly with this? Maybe we should have optional ```mimeType``` for the ```get``` method?

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,156 @@
...
+        'image/jpeg',
+        'PUT',
+        body
+      );
+    }
+
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/get`,
+      'PUT',
+      // Query operation by id.
```
azasypkin: __nit:__ looks like we don't use ```operation``` term anymore :) And maybe entire comment is not useful after all

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  // Private methods.
+  [p.getChannel](channels, selector) {
+    let channelKey;
+
+    if (selector.id) {
+      channelKey = selector.id;
+    } else if (selector.kind || typeof selector === 'string') {
+      // That is optional default I was talking above.
```
azasypkin: __nit:__ remove comment :)

app/js/lib/foxbox/services/base.js
```diff
@@ -0,0 +1,156 @@
...
+    return this[p.net].fetchJSON(
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/get`,
+      'PUT',
+      // Query operation by id.
+      body
+    );
+  }
+
+  // Private methods.
+  [p.getChannel](channels, selector) {
```
azasypkin: __nit:__ maybe small jsdoc would be helpful

app/js/lib/foxbox/services/ip-camera.js
azasypkin: __nit:__ looks like we don't need ```p.adapter```?

app/js/lib/foxbox/services/ip-camera.js
```diff
@@ -0,0 +1,63 @@
...
+
+  get model() {
+    return this[p.model];
+  }
+
+  get name() {
+    return this[p.name];
+  }
+
+  // Public dynamic property getters.
```
azasypkin: __nit:__ comment is not correct

app/js/lib/foxbox/services/light.js
```diff
@@ -0,0 +1,97 @@
...
+const TYPE = 'Light';
+
+const p = Object.freeze({
+  // Public getter only properties.
+  manufacturer: Symbol('manufacturer'),
+  model: Symbol('model'),
+  name: Symbol('name'),
+
+  // Private getter only properties.
+  adapter: Symbol('adapter'),
```
azasypkin: __nit:__ same here, why do we need this property?

app/js/lib/foxbox/services/light.js
```diff
@@ -0,0 +1,97 @@
...
+
+  get name() {
+    return this[p.name];
+  }
+
+  // Public dynamic property getters.
+  isAvailable() {
+    return this.get('available')
+      .then((response) => {
+        for (let id in response) {
```
azasypkin: __nit:__ maybe ```Object.keys().forEach``` will be safer since we don't check that the property is own one? I'd even force [1] :)

[1] http://eslint.org/docs/rules/guard-for-in

azasypkin: Wait a minute, why do we need ```for``` here at all as we return on the first iteration?

app/js/lib/foxbox/services/light.js
```diff
@@ -0,0 +1,97 @@
...
+          return response[id].OnOff === 'On';
+        }
+      });
+  }
+
+  isOn() {
+    return this.get('LightOn')
+      .then((response) => {
+        for (let id in response) {
```
azasypkin: __nit:__ same question and nit here

app/js/lib/foxbox/services/light.js
```diff
@@ -0,0 +1,97 @@
...
+    return this.set('LightOn', value);
+  }
+
+  /**
+   * Turn on the bulb.
+   *
+   * @return {Promise}
+   */
+  turnOn() {
+    return this.set('LightOn', true);
```
azasypkin: __nit:__ maybe it will be better to make ```turn(on)``` as private and call it inside ```turnOn``` and ```turnOff```. Or leave it public if you think it gives more flexibility

app/js/lib/foxbox/services/light.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Turn off the bulb.
+   *
+   * @return {Promise}
+   */
+  turnOff() {
```
azasypkin: __issue:__ same issue here

app/js/lib/foxbox/foxbox.js
```diff
@@ -387,6 +391,9 @@ export default class Foxbox extends Service {
         if (hasNewServices ||
           fetchedServices.length !== storedServices.length) {
           // The state of the services changes.
```
azasypkin: __nit:__ comment doesn't seem to be correct

app/js/lib/foxbox/foxbox.js
```diff
@@ -425,97 +434,54 @@ export default class Foxbox extends Service {
...
-      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/channels/get`,
-      'PUT',
-      payload
-    );
-  }
+    switch (data.adapter) {
+      case 'thinkerbell-adapter':
+      case 'webpush@link.mozilla.org':
+      case 'console@link.mozilla.org':
+        return data;
```
azasypkin: mm, I think we should not return raw data anymore, let's just return ```BaseService``` for all unknown services in ```default:``` branch?

app/js/lib/foxbox/services/ip-camera.js
```diff
@@ -0,0 +1,63 @@
...
+  // Private getter only properties.
+  adapter: Symbol('adapter'),
+});
+
+export default class IpCameraService extends BaseService {
+  constructor(props, config) {
+    super(props, config);
+
+    // Public getter only properties.
+    this[p.manufacturer] = props.properties.manufacturer;
```
azasypkin: __question:__ How do you think, can we move ```manufacturer, mode and name``` to base service?

app/js/views/dev/camera-latest-image.jsx
```diff
@@ -10,36 +10,34 @@ export default class CameraLatestImage extends BaseView {
...
-      this.setState({ hasPreview: true });
-    })
-    .catch((e) => {
-      console.error(
-        'Error occurred while retrieving latest image for camera (id=%s): ',
-        this.props.id,
-        e
-      );
-    });
+        return this.state.service.takeSnapshot();
```
azasypkin: __potential issue:__ iirc I read the react doesn't refresh state immediately, I'd say it would be safe to use ```service``` here directly

app/js/views/dev/camera-latest-image.jsx
```diff
@@ -62,45 +60,9 @@ export default class CameraLatestImage extends BaseView {
       </div>
     );
   }
-
```
azasypkin: \0/

app/js/views/services-list.jsx
```diff
@@ -11,13 +11,7 @@ export default class ServicesList extends React.Component {
     let serviceNodes = userFacingServices.map((service) => (
         <ServicesListItem
           key={service.id}
-          id={service.id}
```
azasypkin: :+1:

azasypkin:
\0/ looks good overall, I've noted few issues, nits and asked few questions. I'm eager to land this PR as early as possible, so tell me when it's ready for the last review round!

app/js/lib/foxbox/services/base.js
gmarty: Yeah that's a good point. I'll change it.

app/js/lib/foxbox/services/base.js
gmarty: I replace it by private variables that are set at instantiation, because getters and setters don't change. So now `hasGetters` and `hasSetters` come with no hidden cost.

app/js/lib/foxbox/services/base.js
gmarty: What do you mean?

app/js/lib/foxbox/services/base.js
gmarty: The lights work correctly. Why do we need to make things more explicit? Does it feel too magical to you?

app/js/lib/foxbox/services/ip-camera.js
gmarty: I wasn't too sure whether it was going to be optional properties, but I think it makes a lot of sense to move them to the base class. I'll default them to an empty string.

app/js/lib/foxbox/services/base.js
azasypkin: I mean, the type of parameter is Object|string and description of the parameters would make it clearer what this ```selector``` is for.

app/js/lib/foxbox/services/ip-camera.js
gmarty: Indeed, `adapter` is not used anywhere.

app/js/lib/foxbox/services/light.js
gmarty: I replace it with:
```js
return this.get('LightOn')
  .then((response) => response[Object.keys(response)[0]].OnOff === 'On');
```

Do you think we need to enforce the `guard-for-in` rule of ESLint? If so, comment on #47.

app/js/lib/foxbox/services/light.js
gmarty: It's hard to say how we'll use this API, but I like the `turn()` one. I think it's clearer than:
```js
if (on) {
  light.turnOn();
} else {
  light.turnOff();
}

File: app/js/views/dev/camera-latest-image.jsx (hunk 24 above)
Comment (gmarty) [inline app/js/views/dev/camera-latest-image.jsx]: Yeah, that's safer.

File: app/js/lib/foxbox/foxbox.js (hunk 22 above)
Comment (gmarty) [inline app/js/lib/foxbox/foxbox.js]: I can't really think of any benefits of instantiating objects for these adapters. It seems to me like a waste of time and memory.

Comment (azasypkin) [inline app/js/lib/foxbox/foxbox.js]: Hmm, maybe I'm missing something, but the code that will be getting list of services - will have list of different types (some of them will be proper BaseService instances and some of them will be raw data with unpredictable structure). Isn't that a big problem?

Comment (gmarty) [inline app/js/lib/foxbox/foxbox.js]: Yes, you're right that's totally inconsistent, but these blacklisted adapters shouldn't be exposed to the user anyway. I remember we had this discussion about whether the filtering of unused adapters should take place in the view or in the lib. You did it in the view but I still think it should happen in the lib.
I suggest to continue doing it in the view for now, but in the future I'd really like to reconsider this option, as long as these adapters won't be used in the view or exposed to the user.
What do you think?

**Comment by gmarty** [comment]
Thanks for this first round. The next version is ready to be reviewed!

In the long term, we need to discuss these items:
* Should `foxbox` lib hold a pool of service wrapper instances?
* Where should the polling happen? In `foxbox` like today, or in the service wrapper classes?
* What about the tags? Should we remove the frontend implementation and use the ones from the box?

File: app/js/lib/foxbox/services/base.js (hunk 11 above)
Comment (azasypkin) [inline app/js/lib/foxbox/services/base.js]: Sorry, I was probably too tired yesterday :) I haven't noticed ```if (getter.kind.type === 'Binary') {``` (that actually I added a while ago) - so yeah, it looks fine for now.

File: app/js/lib/foxbox/services/ip-camera.js [hunk 27]
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  // Public getter only properties.
+  get type() {
+    return TYPE;
+  }
+
+  // Public methods.
+  getImageList() {
+    return this.get('image_list');
```
azasypkin: __funny fact:__ now when camera class is small and clear it's obvious that names of channel kind for image list and latest images look inconsistent :)

app/js/lib/foxbox/services/ip-camera.js
```diff
@@ -0,0 +1,31 @@
...
+
+const TYPE = 'ip-camera';
+
+export default class IpCameraService extends BaseService {
+  constructor(props, config) {
+    super(props, config);
+    Object.seal(this);
+  }
+
+  // Public getter only properties.
```
azasypkin: __question:__ Sorry I still struggle to understand why do we need this comments ```Public getter only properties.``` or ```// Public methods.``` - isn't it clear that ```get``` means getter and name like ```[p.method]``` means private method?

On the related note, I agree that we should group and order getters -> public methods -> private methods. So that reader/developer sees public interface first and only scroll further down if he needs details in private methods.

app/js/lib/foxbox/services/light.js
```diff
@@ -0,0 +1,57 @@
...
+    return this.set('LightOn', value);
+  }
+
+  /**
+   * Turn on the bulb.
+   *
+   * @return {Promise}
+   */
+  turnOn() {
+    return this.set('LightOn', 'On');
```
azasypkin: __question:__ We don't use these methods anywhere yet, maybe let's keep it simple and add new methods only when we have such a need?

__nit:__ In case you decide to leave this method, it would be better to call ```this.turn(true)``` and ```this.turn(false)``` inside so that only one method has this unique and custom ```LightOn``` string.

app/js/lib/foxbox/foxbox.js
```diff
@@ -402,18 +409,20 @@ export default class Foxbox extends Service {
    */
   getServices() {
     return this[p.db].getServices()
-      .then(services => {
-        return services.map(service => service.data);
-      });
+      .then(services => services.map(
+        service => this[p.getServiceInstance](service.data)
```
azasypkin: __nit:__ round brackets around parameters in arrow functions

app/js/lib/foxbox/foxbox.js
```diff
@@ -402,18 +409,20 @@ export default class Foxbox extends Service {
...
 
   getTags() {
     return this[p.db].getTags.apply(this[p.db], arguments);
   }
 
+  // @todo If service doesn't exist in the DB, fetch it from the box.
   getService() {
     // Get data from the DB so we get the attributes, the state and the tags.
-    return this[p.db].getService.apply(this[p.db], arguments);
+    return this[p.db].getService.apply(this[p.db], arguments)
```
azasypkin: __nit:__ round brackets around parameters in arrow functions

app/js/lib/foxbox/foxbox.js
```diff
@@ -425,97 +434,49 @@ export default class Foxbox extends Service {
...
-          };
-        });
-      });
+      `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`
+      )
+      .then((services) => services.map((service) => {
+        // Some service don't have name, but can have product_name instead,
+        // eventually we should get rid of this line and use unified service
+        // label field provided by taxonomy.
+        service.properties.name = service.properties.name ||
```
azasypkin: __optional nit:__ it feels like BaseService constructor now is the better place for this workaround, but up to you.

azasypkin:
Okay, now looks really good! I've left few more nits and several questions, and Travis is failing.

r+ with green travis!

app/js/lib/foxbox/foxbox.js
gmarty: Yes, I was thinking exactly the same.

app/js/lib/foxbox/services/base.js
azasypkin: Ok, works for me.

app/js/lib/foxbox/services/light.js
azasypkin: Ok, added request for ```guard-for-in``` in #47

app/js/lib/foxbox/foxbox.js
azasypkin: > I suggest to continue doing it in the view for now, but in the future I'd really like to reconsider this option, as long as these adapters won't be used in the view or exposed to the user.
What do you think?

My idea was to pass "selector" to the ```getServices``` so that it queries only services we need to display (let's say user facing service should have particular tag - ```UI``` or something), other views (like recipe) query other sets of services (in our case it's even another request - getters and setters) depending on the view logic. So filtering is done for us by server side and we use much less traffic when fetching only what we need.

But yeah, we need to discuss it further.

azasypkin:
> In the long term, we need to discuss these items:

Just my preliminary thoughts. 

> Should foxbox lib hold a pool of service wrapper instances?

I believe yes.

> Where should the polling happen? In foxbox like today, or in the service wrapper classes?

I'd leave it as is in the hope that it will die soon - once we get websockets or event source. But what I'd like to do right now is to make it much-much less aggressive - my Firefox is dying when I have dev console opened for our app :/ 

> What about the tags? Should we remove the frontend implementation and use the ones from the box?

You mean the one that relies on IndexedDB? If yes, I think so and we need to make UI fully functional. Nobody uses tags right now, but it's really critical for Recipe pages (we query getters and setters and can only distinguish them by tags, so we need to have way to tag services/getters/setters). See #98 and #90. But we need UX/VD input on this, I'd ask Juwei to help.

app/js/lib/foxbox/services/ip-camera.js
gmarty: It's more for code organisation so that we have a consistent ordering across classes.
If you think that's useless and we're going to keep this ordering, then I can remove them.

app/js/lib/foxbox/services/light.js
gmarty: OK, you're right, let's remove them for now. We'll see if we need them in the future.

app/js/lib/foxbox/services/ip-camera.js
azasypkin: Yeah, I'd make just a strict rule to group getters, public and private methods and order them appropriately.

PR #75, opened 2016-04-13
Remove Cordova code

michielbdejong:
r? @gmarty


gmarty:
Looks good to me. r+
But maybe we want to keep these files in a branch in case we want to reactivate it later.

michielbdejong:
Yes, the Cordova code will live on in https://github.com/michielbdejong/serverless-link-client.

PR #76, opened 2016-04-13
Fix the fill-body style

gmarty:


gmarty:
@azasypkin: r? please.

app/css/app.css
```diff
@@ -100,6 +100,9 @@ select {
 }
 
 .app-view__fill-body {
+  position: absolute;
```
azasypkin: I think here is the Chrome issue that has been fixed [1] and reached my dev version, but still on its way to your beta.

I've seen two workarounds on the bug:

* add ```display: flex``` to the ```.app-view__body``` and replace ```height: 100%``` here with ```flex-grow: 1;```;
* same solution as you did.

Honestly I like the flex-based workaround, but you choose :) But in any case please add comment with the link to Chromium bug so that we can get rid of this hack later.

[1] https://bugs.chromium.org/p/chromium/issues/detail?id=341310

azasypkin:
Thanks, __r+__ with any of the solutions I mentioned in https://github.com/fxbox/app/pull/76/files#r59535717

gmarty:
Thanks for the investigation. If it's a bug I'm not sure if it's worth being fixed.
But maybe we should switch to flexbox everywhere for the sake of consistency.

gmarty:
The bug has been fixed in Chrome. Closing.

PR #79, opened 2016-04-13
Adapt recipes UI to new Taxonomy API (fixes #68). r=gmarty

### LINKED ISSUE #68: [Recipies] It's impossible to add a new recipe

azasypkin:


app/js/lib/foxbox/recipes.js
```diff
@@ -6,10 +6,35 @@
...
   settings: Symbol('settings'),
-  net: Symbol('net')
+  net: Symbol('net'),
+
+  // Recipe private properties.
+  service: Symbol('service')
 });
 
-export default class Recipe {
+export class Recipe {
```
azasypkin: Just first draft of the ```Recipe``` object just to have the same "view/projection" that is used inside app views, later we can make it much smarter.

app/js/lib/foxbox/recipes.js
```diff
@@ -6,10 +6,35 @@
...
+
+    this[p.service] = service;
+  }
+
+  get id() {
+    return this[p.service].id;
+  }
+
+  get label() {
+    return this[p.service].source && this[p.service].source.name;
```
azasypkin: As it turned out we can have recipe without source (if something went wrong when we were creating a recipe). Likely we'll get rid of it later, but right now it saves a lot of development time - while experimenting with the taxonomy requests there is a high chance to create broken recipe :)

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
-            { id: `${recipe.id}/get_enabled` }
-          ];
-
-          return this[p.net].fetchJSON(
-            `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/` +
-            'channels/get',
-            'PUT',
-            payload
-          );
+    let recipeServicesPromise = this[p.net].fetchJSON(
```
azasypkin: So we need only recipes to display in the list

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
+          setEnabled: enabledSetterId,
+          remove: removeSetterId,
+          status: null,
+          source: null
+        };
       });
-  }
+    });
 
-  getServicesWithGetters() {
```
azasypkin: Refactoring is my only way to quickly get something done, sorry for the big changeset :/

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
+
+          return selectors;
+        },
+        { enabledSelectors: [], sourceSelectors: [] }
+      );
+
+      const getterURL = `${this[p.net].origin}/api/` +
+        `v${this[p.settings].apiVersion}/channels/get`;
+
+      return Promise.all([
```
azasypkin: Getting all statuses and sources (to get name of the recipe) for all recipes in two requests :) Actually we can do it even in one, but it didn't work for me - need to discuss.

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
-                  value: 'service:clock@link.mozilla.org,CurrentTimeOfDay,eq,' +
-                  '06:00pm'
-                }
-              ];
-              break;
-
-            default:
-              //service.getters = service.getters;
-              break;
+          if (sourceResponse && sourceResponse.String) {
```
azasypkin: Just basic guards for the cases I've mentioned above, once we have stable API we can more intelligent and generic checks.

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
+          return new Recipe(service);
+        });
+      });
+    });
   }
 
-  getServicesWithSetters() {
+  getGetters() {
+    // Currently we support only Clock as a trigger.
+    const supportedKinds = ['CurrentTimeOfDay'];
```
azasypkin: If you don't mind, for the demo I'd rather indicate services that we support, instead of filtering out unsupported ones, I've just found it a bit easier to add new supported services. I guess the next one will be motion sensor

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
+            options.push(...[{
+              label: 'in the morning',
+              // 08:00 AM, 8 * 60 * 60 = 28800 seconds from 00:00.
+              value: { Geq: { Duration: 28800 } }
+            }, {
+              label: 'in the afternoon',
+              // 02:00 PM, 14 * 60 * 60 = 50400 seconds from 00:00.
+              value: { Geq: { Duration: 50400 } }
+            }, {
+              label: 'in the evening',
```
azasypkin: Locally I test it with the following additional value option:
```
const twoMinutesFromNow = new Date();
twoMinutesFromNow.setMinutes(twoMinutesFromNow.getMinutes() + 2);
twoMinutesFromNow.setSeconds(0);
......
{
  label: `at ${twoMinutesFromNow.toLocaleTimeString()}`,
  // 06:00 PM, 18 * 60 * 60 = 64800 seconds from 00:00.
  value: { Geq: {
    Duration: twoMinutesFromNow.getHours() * 60 * 60 +
      twoMinutesFromNow.getMinutes() * 60
  }}
}
```

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
+          kind: getter.kind,
+          name: name || getter.adapter,
+          options
+        };
+      });
+    });
+  }
+
+  getSetters() {
+    // Currently we support only Camera and TTS as actuator.
```
azasypkin: We'll need to add lights I suppose

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
+    .then(setters => {
+      return setters.map(setter => {
+        let name, options = [];
+
+        // Check for the complex extension kind
+        if (typeof setter.kind === 'object') {
+
+          // WORKAROUND: waiting for the Taxonomy fix so that we have "type" for
+          // all places.
+          setter.kind.type = setter.kind.typ;
```
azasypkin: Diiiiiirty hack, waiting for Yoric to land fix, we'll get rid of it as soon as it lands

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
+            }, {
+              label: '"Good afternoon!"',
+              value: { String: '"Good afternoon!"' }
+            }, {
+              label: '"Good evening!"',
+              value: { String: '"Good evening!"' }
+            }]);
+          }
+        } else if (setter.kind === 'TakeSnapshot') {
+          name = 'Camera';
```
azasypkin: Here is one issue I'm going to fix in the next PR - I'd rely on tags that are assosiated with this setter to get name (```Kitchen```, ```Front Door``` etc.)

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
-                }
-              ],
-              kind: {
-                kind: kindGet
-              },
-              range: {
-                [rangeKeyGet]: rangeValGet
-              }
+              source: [{ id: getter.id }],
+              kind: getter.kind,
```
azasypkin: Here is the reason why I store ```kind``` cause it can be both String or Object.

azasypkin:
r? @gmarty 

How does it look to you?

app/js/lib/foxbox/recipes.js
gmarty: You did well, it was all messy!

app/js/lib/foxbox/recipes.js
gmarty: Yes, and it's in line with the view (like `isUserFacingService()` in `views/services-list.jsx`)

app/js/lib/foxbox/recipes.js
gmarty: Yes, this and the open zwave, but let's do it in a subsequent patch.

app/js/lib/foxbox/recipes.js
```diff
@@ -6,10 +6,35 @@
 const p = Object.freeze({
   // Private properties.
   settings: Symbol('settings'),
-  net: Symbol('net')
+  net: Symbol('net'),
+
+  // Recipe private properties.
+  service: Symbol('service')
```
gmarty: We can start adding comma at the end of multilines.

app/js/lib/foxbox/recipes.js
gmarty: Don't you prefer to move this class to its own file?

app/js/views/themes-new.jsx
```diff
@@ -12,120 +12,118 @@ export default class ThemesNew extends BaseView {
...
   }
 
-  handleServiceWithSetChange(evt) {
-    const serviceWithSet = evt.target.value;
-    if (serviceWithSet !== '') {
-      this.setState({ serviceWithSet });
+  onSetterSelected(evt) {
+    const selectedSetterIndex = evt.target.value;
+    if (selectedSetterIndex) {
+      this.setState({ selectedSetterIndex: +selectedSetterIndex });
```
gmarty: Is the `+` in front of `selectedSetterIndex` to cast it to a Number?
If so, I prefer this to be made in the `const` assignation. Maybe also use something more noticeable (like `Number(selectedSetterIndex)` or `parseInt(selectedSetterIndex)`).

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
    * @return {Promise}
    */
-  add(recipe) {
-    const [idGet, kindGet, rangeKeyGet, rangeValGet] = recipe.getter.split(',');
-    const [idSet, kindSet, rangeKeySet, rangeValSet] = recipe.setter.split(',');
-
-    recipe = {
+  add({ recipeName, getter, getterValue, setter, setterValue }) {
+    const recipe = {
+      name: recipeName,
```
gmarty: Can you rename the param `recipeName` to `name` and you can use Object literal property value shorthand.

app/js/views/themes-new.jsx
```diff
@@ -146,104 +144,125 @@ export default class ThemesNew extends BaseView {
...
+    const optionNodes = this.state.getters.map((getter, index) => (
+      <option key={index} value={index}>{getter.name}</option>
     ));
-    let serviceSelector = (
-      <select value={this.state.serviceWithGet}
-              onChange={this.handleServiceWithGetChange}
+
+    return (
+      <select value={this.state.selectedGetterIndex}
+              onChange={this.onGetterSelected}
```
gmarty: nit: In other JSX files, event listeners are called on the model of `handle...()`. Do you prefer the `on...()` model?

app/js/lib/foxbox/recipes.js
```diff
@@ -24,232 +49,222 @@ export default class Recipe {
...
-
-            default:
-              service.name = service.adapter;
-              break;
+    // Fetch recipe enabled statuses and sources.
+    return recipeServicesPromise.then(services => {
+      // Fetch their respective enabled status.
+      const { enabledSelectors, sourceSelectors } = services.reduce(
+        (selectors, service) => {
+
```
gmarty: nit: unneeded line break.

app/js/lib/foxbox/recipes.js
gmarty: Why do you assign the promise to a variable instead of the usual cascading starting by `return this[p.net].fetchJSON(`?

app/js/lib/foxbox/recipes.js
azasypkin: Yep, will add

app/js/lib/foxbox/recipes.js
azasypkin: I was just not sure about this one, but I can move it to ```lib/foxbox/recipe.js```?

app/js/lib/foxbox/recipes.js
azasypkin: Ah it's just leftover - I was refactoring step-by-step - will merge them (but eventually we may need to split them in separate methods to read it easier)

gmarty:
I love this recipe v2.0. We're getting a more complete and readable code. Let's r+ and land this.

app/js/lib/foxbox/recipes.js
azasypkin: Hmm yep!

app/js/lib/foxbox/recipes.js
azasypkin: yep, sure

app/js/views/themes-new.jsx
azasypkin: > Is the + in front of selectedSetterIndex to cast it to a Number?
Exactly.

> If so, I prefer this to be made in the const assignation. Maybe also use something more noticeable (like Number(selectedSetterIndex) or parseInt(selectedSetterIndex)).

Ok, makes sense

app/js/lib/foxbox/recipes.js
azasypkin: Mm just realised that I use ```[p.service]``` inside both ```Recipe``` and ```Recipes```. So I see three options:
* Leave it as is;
* Move ```Recipe``` to a separate file and expose public method like ```raw()``` that will return internal data (```this[p.service]``` now);
* Move ```Recipe``` to a separate file, leave ```[p.service]``` private, but use something like ```service: Symbol.for('recipe-raw')``` so that it can be used by both ```Recipe``` and ```Recipes```.

What do you prefer?

gmarty: Oh I haven't thought of that. I don't really want to expose it as raw data. Maybe we should leave it here then, hopefully that won't impact badly unit testing of these classes.

PR #87, opened 2016-04-14
Add support for connected lights in the recipes. r=azasypkin

gmarty:


gmarty:
I reformatted a few things in the `getSetters()` method mainly to reduce the git diff when we add or move around items in the future.
I noticed the `then(` are not properly indented, but fixing it would make the diff more complicated so I didn't try to fix them.

app/js/lib/foxbox/recipes.js
```diff
@@ -198,36 +199,53 @@ export default class Recipes {
...
+              },
+            ]);
           }
         } else if (setter.kind === 'TakeSnapshot') {
-          name = 'Camera';
+          name = 'camera';
           options.push({
-              label: 'take a picture',
-              value: { 'Unit': null }
+            label: 'takes a picture',
```
gmarty: Because we use the labels to make phrases, I had to remove the upper case and make sure it's still proper English when concatenated.

gmarty:
@azasypkin: r? please.

app/js/lib/foxbox/recipes.js
azasypkin: Ah yeah, right

azasypkin:
lgtm, thanks! r+

app/js/lib/foxbox/recipes.js
```diff
@@ -198,36 +199,53 @@ export default class Recipes {
...
+            value: { 'Unit': null }
           });
+        } else if (setter.kind === 'LightOn') {
+          name = 'light';
```
azasypkin: We need to figure out tomorrow how to tag them - so that when we have several lights/motion sensors/cameras we can differentiate them...

app/js/lib/foxbox/recipes.js
```diff
@@ -198,36 +199,53 @@ export default class Recipes {
...
+          options.push(...[
+            {
+              label: 'gets turned on',
+              value: { 'OnOff': 'On' }
```
azasypkin: __nit:__ there is no need in quotes for ```OnOff``` so just ```{ OnOff: 'On' }``` and same below

app/js/lib/foxbox/recipes.js
gmarty: Hopefully tags will help here. We'll take a look after https://github.com/fxbox/app/pull/73 lands.

PR #88, opened 2016-04-14
Webpush click and notificating via postMessage

arcturus:
@gmarty r?

### DIFF
--- a/app/hookSW.js
+++ b/app/hookSW.js
@@ -19,9 +19,12 @@ self.addEventListener('push', function(evt) {
     notifyClient(obj)
       .then(() => {
         showNotification(obj, evt);
+      })
+      .catch((error) => {
+        console.error('Error processing notification: ', error);
       });
   } else {
-    console.log('Discarding notification: ' + JSON.stringify(obj));
+    console.error('Notification doesnt contain a body, ignoring it: ');
   }
 });
 
@@ -33,7 +36,8 @@ self.addEventListener('notificationclick', function(evt) {
   evt.notification.close();
   evt.waitUntil(clients.matchAll({
     type: "window"
-  }).then(function(clients) {
+  })
+  .then(function(clients) {
     clients.forEach((client) => {
       if ('focus' in client) {
         client.focus();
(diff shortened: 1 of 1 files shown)


app/hookSW.js
```diff
@@ -15,24 +15,61 @@ self.addEventListener('push', function(evt) {
   console.log('Push received ', JSON.stringify(evt.data));
   var obj = evt.data ? evt.data.json() : {};
 
-  if (obj && obj.body) {
-    showNotification(obj, evt);
+  if (obj && obj.message) {
+    notifyClient(obj)
+      .then(() => {
+        showNotification(obj, evt);
+      });
```
gmarty: Would be good to catch the promise here in case anything goes wrong.

app/hookSW.js
```diff
@@ -15,24 +15,61 @@ self.addEventListener('push', function(evt) {
...
   } else {
-    console.log('Discarding notification');
+    console.log('Discarding notification: ' + JSON.stringify(obj));
```
gmarty: `console.error`, also maybe an easier message like `Notification doesn't contain a body, ignoring it`.

app/hookSW.js
```diff
@@ -15,24 +15,61 @@ self.addEventListener('push', function(evt) {
...
 
 self.addEventListener('pushsubscriptionchange', function(evt) {
   console.log('Got subscription change event ', evt);
 });
 
+self.addEventListener('notificationclick', function(evt) {
+  evt.notification.close();
+  evt.waitUntil(clients.matchAll({
+    type: "window"
+  }).then(function(clients) {
```
gmarty: nit: Line break before `.then(` please.

app/hookSW.js
```diff
@@ -15,24 +15,61 @@ self.addEventListener('push', function(evt) {
...
+        client.focus();
+      }
+    });
+  }));
+});
+
+/**
+ * Notify the foxlink library of the message received via push.
+ * This will notify all the windows/tabs opened with the app.
+ * @param {object} obj Payload coming from the push notification
```
gmarty: nit1: We usually add a line break between the description and the JSDoc annotations.
nit2: We use the [JSDoc annotations as implementend in Closure Compile](https://developers.google.com/closure/compiler/docs/js-for-compiler#types), so it's `Object` (capitalised).

app/hookSW.js
```diff
@@ -15,24 +15,61 @@ self.addEventListener('push', function(evt) {
...
+      clients.forEach((client) => {
+        client.postMessage(obj);
+      });
+    });
+}
+
+/**
+ * Displays a notification based on the data coming from the
+ * push.
+ * @param {object} obj Payload coming from the push notification
```
gmarty: see above.

gmarty:
I suggested a few minor changes, but r+!

PR #89, opened 2016-04-15
Introduce Motion Sensor support. r=gmarty

azasypkin:


app/js/lib/foxbox/foxbox.js
```diff
@@ -368,7 +368,15 @@ export default class Foxbox extends Service {
   getServices() {
     return this[p.db].getServices()
       .then(services => {
-        return services.map(service => service.data);
+        return services.map(service => {
+          // Some service don't have name, but can have product_name instead,
+          // eventually we should get rid of this line and use unified service
+          // label field provided by taxonomy.
+          service.data.properties.name = service.data.properties.name ||
```
azasypkin: There is a high chance (at the moment I think it's 100% chance) that OpenZWave devices won't have name, but there is a much higher chance that they we'll have product name. It's harder to change at the server side before demo so we agreed with @julienw to fix this on the client side for now.

app/js/lib/foxbox/foxbox.js
```diff
@@ -368,7 +368,15 @@ export default class Foxbox extends Service {
...
+            service.data.properties['product_name'];
```
azasypkin: EsLint is not happy with ```product_name```, so I had to put it in quotes.

app/js/views/services-list.jsx
```diff
@@ -27,9 +27,17 @@ export default class ServicesList extends React.Component {
   }
 
   isUserFacingService(service) {
+    // If service doesn't have any getters and setters, there is no need to
+    // display it to the user.
+    if (!Object.keys(service.getters).length &&
```
azasypkin: It's also short term fix to filter out useless services (like OpenZWave controller), later one I think we'll be requesting getters and setters instead and group them if needed. Maybe.

app/js/lib/foxbox/foxbox.js
```diff
@@ -500,6 +500,12 @@ export default class Foxbox extends Service {
       `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
       .then((services) => {
         return services.map((service) => {
+          // Some service don't have name, but can have product_name instead,
+          // eventually we should get rid of this line and use unified service
+          // label field provided by taxonomy.
+          service.properties.name = service.properties.name ||
+            service.properties['product_name'];
```
azasypkin: EsLint is not happy with product_name, so I had to put it in quotes.

app/js/lib/foxbox/foxbox.js
```diff
@@ -500,6 +500,12 @@ export default class Foxbox extends Service {
       `${this[p.net].origin}/api/v${this[p.settings].apiVersion}/services`)
       .then((services) => {
         return services.map((service) => {
+          // Some service don't have name, but can have product_name instead,
+          // eventually we should get rid of this line and use unified service
+          // label field provided by taxonomy.
+          service.properties.name = service.properties.name ||
```
azasypkin: There is a high chance (at the moment I think it's 100% chance) that OpenZWave devices won't have name, but there is a much higher chance that they have product_name. We agreed with @julienw that it will be easier to fix this on client side for the demo.

app/js/lib/foxbox/recipes.js
```diff
@@ -135,8 +135,8 @@ export default class Recipes {
   }
 
   getGetters() {
-    // Currently we support only Clock as a trigger.
-    const supportedKinds = ['CurrentTimeOfDay'];
+    // Currently we support only Clock and Motion Sensor as triggers.
+    const supportedKinds = ['CurrentTimeOfDay', 'OpenClosed'];
```
azasypkin: For now we consider ```OpenClosed``` as Motion Sensor, but later we should introduce dedicated ChannelKind

app/js/lib/foxbox/foxbox.js
gmarty: Is it a bug in ESLint?

azasypkin: I thought it forbids underscores in the names, but didn't pay attention - let me double check

gmarty:
Looks good to me. Let see if we can fix the ESLint error and then let's land it! r+

app/js/lib/foxbox/foxbox.js
azasypkin: Mm, sorry, gulp lint looks happy, it just was my Webstorm who warned me :/

azasypkin:
> Looks good to me. Let see if we can fix the ESLint error and then let's land it! r+

Thanks! I've fixed ```['product_name']``` (it was more my error than EsLint :)). I'll land this once I confirm that it works for real motion sensor - in the progress :)

azasypkin:
Okay, everyone is happy, landing...

PR #91, opened 2016-04-15
Add support for custom demo/dev views. r=gmarty

azasypkin:


app/js/controllers/dev.js
```diff
@@ -0,0 +1,21 @@
+import React from 'components/react';
+import ReactDOM from 'components/react-dom';
+import { Controller } from 'components/mvc';
+
+import CameraLatestImageView from 'js/views/dev/camera-latest-image';
+
+export default class DevController extends Controller {
```
azasypkin: Didn't want to pollute existing views/controllers with such temporal stuff and think that we may need more such views in the future, so here is the idea with Dev controller that can serve various dev/demo related views.

How does it look to you?

app/js/views/dev/camera-latest-image.jsx
```diff
@@ -0,0 +1,106 @@
...
+  renderBody() {
+    let cameraControlsClass = 'app-view__fill-body camera-controls';
+    if (this.state.hasPreview) {
+      cameraControlsClass += ' camera-controls--has-preview';
+    }
+
+    return (
+      <div className={cameraControlsClass}>
+        <img ref="snapshotPreview"
+             style={{ flexGrow: 0 }}
```
azasypkin: There no camera controls here, so it's the easiest fix to keep aspect ratio. Eventually we should just fix camera service view, I just don't want to spend time on this at the moment.

app/js/controllers/dev.js
gmarty: I just discussed with @arcturus and we think that redirecting the app to specific routes when we get a message from a web push notification is something that's going to stay.
We should rename `DevController` to a better name (like `ServiceGetterController` or `GetterController` or anything else).

The route too should reflect that. I was thinking a sub route in service like `http://localhost:8000/#service/service:18db9b78f188-11e5-9ce9-3ca9f47e346c@link.mozilla.org/latest-image`, but I leave it to you as long as the URL structure is logical and makes sense.

app/js/controllers/main.js
```diff
@@ -22,7 +23,8 @@ export default class MainController extends RoutingController {
       'services': new ServicesController(options),
       'services/(.+)': new ServiceController(options),
       'themes': themesController,
-      'themes/(.+)': themesController
+      'themes/(.+)': themesController,
+      'dev/(.+)/(.+)': new DevController(options),
```
gmarty: See above.

app/js/views/dev/camera-latest-image.jsx
gmarty: That works for me.

app/js/views/dev/camera-latest-image.jsx
```diff
@@ -0,0 +1,106 @@
...
+             alt={'Snapshot preview'} className="camera-controls__preview" />
```
gmarty: You can just do `alt="Snapshot preview"`.

gmarty:
The code looks good to me. We need to find a better route and naming but r+. Thanks!

app/js/controllers/dev.js
azasypkin: TLDR; Well that doesn't really sound right from pov of taxonomy architecture we'd like to work with and cases that will be related to web notification, but it's not easy to articulate :)

It doesn't feel like we need something related to just data returned from the getter, but more like an independent view with specific purpose - in this case it's photo that has been made by the camera that has been triggered by specific recipe (maybe this view should say that this photo has been taken when door/closed sensor was activated). Next we can have notification that should redirect user to a view that will use more than one getter and ```GetterController``` and URL structure won't scale, I'm not sure if can generalize on this only use case right now. But it requires more time to think about it to make it right.

What should ```latest-image``` in this URL be - is it globally unique getter id (in this case id of the service is redundant, but we still need it to get service name)? Tag? ChannelKind? Some developer-friendly-arbitary-string?

But we need it today/now, so I'm thinking about compromise at the moment...

arcturus: I understand the point, and basically as a first hacky approach for having a demo for the next 24 hours could be valid. Period.
TLDR: We all agree, this is a temporary solution to show case what we can do with the app. We need more thinking.

As you comment, we need to sit down and think how we want to organise not just the notifications, but the communication box -> user. Simplifying it just to a controller or existing controller definitely won't scale much, but we need as well to define what kind of information we will get.

Right now we can have predefined static strings, already configured in the rules, again doesn't seem to me really powerful. Or better said, we are limiting ourselves a lot :).

We need to sit with @aosmond and @Yoric to define this kind of communication, or future communications :)

azasypkin: > I understand the point, and basically as a first hacky approach for having a demo for the next 24 hours could be valid. Period.
TLDR: We all agree, this is a temporary solution to show case what we can do with the app. We need more thinking.

True, I'm just trying to understand why do we want to mimic kind of final and generic approach with "ServiceGetterController or GetterController" and not to go with temporal anything temporal that I have now

PR #95, opened 2016-04-15
Add a temporary method to create recipes for the demo

gmarty:


arcturus:
r+ landing

app/js/lib/foxbox/recipes.js
```diff
@@ -136,7 +136,7 @@ export default class Recipes {
 
   getGetters() {
     // Currently we support only Clock and Motion Sensor as triggers.
-    const supportedKinds = ['CurrentTimeOfDay', 'OpenClosed'];
+    const supportedKinds = ['CurrentTimeOfDay', 'OpenClosed', 'LightOn'];
```
azasypkin: @gmarty mm does it mean we'll have getter without any options/values in ```IF``` section in the recipe page?

gmarty: I don't really know actually. I'm afraid to clutter the UI if we expose every single setter and getter and I can't really find any values in being able to query the bulb power state.
But maybe I'm wrong and we need to expose them all.
If you think we need all setter and all getter, let's update the code. Otherwise, we'll wait on an improved UI that works better than drop down list.

PR #96, opened 2016-04-15
Issue #9 - Do not block UI on box discovery. Do not default to localhost. Disabl…

ferjm:
…e login when no box found. r=gmarty

### DIFF
--- a/app/css/views/user-login.css
+++ b/app/css/views/user-login.css
@@ -19,6 +19,10 @@
   border: none;
 }
 
+.user-login__login-button:disabled {
+  background: #c1d9d8;
+}
+
 .user-login__box-selector {
   margin: 1rem;
   padding: 1rem;
--- a/app/js/lib/foxbox/README.md
+++ b/app/js/lib/foxbox/README.md
@@ -52,3 +52,8 @@ is disconnected from it.
 ### `service-state-change`
 
 This event is emitted when a service states change.
+
+### `box-online`
+
+This event is emitted once a box is discovered on the local network or every
+time we get a response from a ping to a previously discovered box url (local or remote).
--- a/app/js/lib/foxbox/foxbox.js
+++ b/app/js/lib/foxbox/foxbox.js
@@ -52,7 +52,9 @@ export default class Foxbox extends Service {
     // Private properties.
     this[p.settings] = new Settings();
     this[p.db] = new Db();
-    this[p.net] = new Network(this[p.settings]);
+    this[p.net] = new Network(this[p.settings], foxboxOnline => {
+      this._dispatchEvent('box-online', foxboxOnline);
+    });
     this[p.boxes] = Object.freeze([]);
     this[p.isPollingEnabled] = false;
     this[p.nextPollTimeout] = null;
@@ -70,28 +72,30 @@ export default class Foxbox extends Service {
   init() {
     window.foxbox = this;
 
+    // No need to block the UI on the discovery process.
+    // Once we discover a box we can connect to, we will start
+    // polling and triggering box-online events with a boolean
+    // indicating if we have access to box or not.
+    this._initDiscovery().then(() => {
+      return this[p.net].init();
+    })
+    .then(() => {
+      // Start polling.
+      this[p.settings].on('pollingEnabled', () => {
+        this.togglePolling(this[p.settings].pollingEnabled);
+      });
+      this.togglePolling(this[p.settings].pollingEnabled);
+
+      this.recipes = new Recipes({
+        settings: this[p.settings],
+        net: this[p.net]
+      });
+    });
+
     return this._initUserSession()
-      .then(() => {
-        return this._initDiscovery();
-      })
-      .then(() => {
-        return this[p.net].init();
-      })
       .then(() => {
         // The DB is only initialised if there's no redirection to the box.
         return this[p.db].init();
-      })
-      .then(() => {
-        // Start polling.
-        this[p.settings].on('pollingEnabled', () => {
-          this.togglePolling(this[p.settings].pollingEnabled);
-        });
-        this.togglePolling(this[p.settings].pollingEnabled);
-
-        this.recipes = new Recipes({
-          settings: this[p.settings],
-          net: this[p.net]
-        });
       });
   }
 
@@ -118,7 +122,7 @@ export default class Foxbox extends Service {
   /**
    * Get the URL of the box using the registration server.
    * If it fails, we fallback to the previously set hostname.
-   * It there isn't, it falls back to localhost.
+   * It there isn't, we schedule a retry.
    *
    * @returns {Promise}
    * @private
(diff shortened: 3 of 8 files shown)


ferjm:
I think this would also help with the mixed content blocked that was causing the https -> http redirection on github pages.

app/js/lib/foxbox/foxbox.js
```diff
@@ -64,28 +64,27 @@ export default class Foxbox extends Service {
...
+      this.togglePolling(this[p.settings].pollingEnabled);
+
+      this.recipes = new Recipes({
+        settings: this[p.settings],
+        net: this[p.net]
+      });
+    });
+
     return this._initUserSession()
       .then(() => {
```
gmarty: Can you give more details about why moving the init user session there?
This methods contains the code that redirects the UI and needs to happen earlier, before setting polling and recipes at least.

app/js/lib/foxbox/foxbox.js
```diff
@@ -117,49 +116,66 @@ export default class Foxbox extends Service {
...
-            this.selectBox();
-          }
-          resolve();
-        })
-        .catch(() => {
-          // When something goes wrong, we still want to resolve the promise so
-          // that a hostname set previously is reused.
-          resolve();
-        });
+    return new Promise(resolve => {
```
gmarty: nit: wrap arrow function argument in parens.

app/js/lib/foxbox/foxbox.js
```diff
@@ -117,49 +116,66 @@ export default class Foxbox extends Service {
...
+                    return Object.freeze(box);
+                })
+            );
+
+            // If the registration server didn't give us any info and
+            // we have no record of previous registrations, we schedule
+            // a retry.
+            if (!this[p.boxes].length &&
+                !this[p.settings]._localOrigin &&
+                !this[p.settings]._remoteOrigin) {
```
gmarty: I thought connecting via the tunnel was optional. Is it still the case?

app/js/lib/foxbox/settings.js
```diff
@@ -89,8 +87,10 @@ export default class Settings extends Model {
 
   set localOrigin(origin) {
     this._localOrigin = origin ?
-        (new URL(origin)).origin : DEFAULT_LOCAL_ORIGIN;
-    storage.setItem(`${PREFIX}localOrigin`, this._localOrigin);
+        (new URL(origin)).origin : undefined;
+    if (this._localOrigin) {
+      storage.setItem(`${PREFIX}localOrigin`, this._localOrigin);
+    }
```
gmarty: If the value was previously set in the local storage but override by a wrong one, should we delete the one from the local storage too?

gmarty:
I still have questions before landing this patch and I also would like to get another pair of eyes on it.
@azasypkin can you take a look, please?

azasypkin:
> @azasypkin can you take a look, please?

Sure, looking

ferjm:
@azasypkin hold on with that review, please. I'm doing some changes to the PR.

azasypkin:
> @azasypkin hold on with that review, please. I'm doing some changes to the PR.

Okay, thanks for the note!

app/js/lib/foxbox/foxbox.js
ferjm: initUserSession does two things:

1. Check if we already have a session.
2. Check if we come from a redirection.

Both things should be independent of the box discovery (and polling) process. The UI shouldn't be blocked by discovering or connecting to the box. Instead we should just inform the user about the discovery and online/offline status of the box (i.e. enabling/disabling the login button and probably later with another visual indicator on the main screen). So far we were avoiding to block the UI defaulting to localhost:3000 when the discovery process failed, but that was wrong and pretty confusing cause once we were setting that default we were stopping trying to reach out the real box url. This is specially annoying now that localhost:3000 does not work any more.

app/js/lib/foxbox/foxbox.js
ferjm: Yes, that's still the case.

app/js/lib/foxbox/settings.js
ferjm: I'm not sure that I understand what you mean here.

Are you suggesting the following scenario?

1. Discover box with url X.
2. Start using box with url X.
3. Discover that box X changed to url Y (note that this requires resetting the box)
4. Fail to use Y connection
5. Go back to url X.

ferjm:
@azasypkin I think this is ready for review. Thanks!

app/js/lib/foxbox/foxbox.js
ferjm: Ok, I'll change it after @azasypkin review.

app/css/views/user-login.css
```diff
@@ -19,6 +19,10 @@
   border: none;
 }
 
+.user-login__login-button:disabled {
+  background: #C1D9D8;
```
azasypkin: __nit:__ please use lower case for the hex colors.

app/js/lib/foxbox/foxbox.js
```diff
@@ -123,49 +127,66 @@ export default class Foxbox extends Service {
    * @returns {Promise}
    * @private
    */
-  _initDiscovery() {
+  _initDiscovery(timeout) {
```
azasypkin: __nit:__ maybe ```timeout = 0``` since you don't pass it when you call this method for the first time

app/js/lib/foxbox/foxbox.js
```diff
@@ -123,49 +127,66 @@ export default class Foxbox extends Service {
...
+              this.selectBox();
+            }
+            console.log('Box found');
+            this._dispatchEvent('box-online', true);
+            resolve();
+          })
+          .catch(() => {
+            // Default to a previously stored box registration.
+            if (this[p.settings]._localOrigin ||
+               this[p.settings]._remoteOrigin) {
```
azasypkin: __issue__ That's may be funny but I don't see where we define ```_remoteOrigin``` is it ```_tunnelOrigin```?

app/js/lib/foxbox/foxbox.js
```diff
@@ -123,49 +127,66 @@ export default class Foxbox extends Service {
...
-          // When something goes wrong, we still want to resolve the promise so
-          // that a hostname set previously is reused.
-          resolve();
-        });
+    return new Promise(resolve => {
+      setTimeout(() => {
+        this[p.net].fetchJSON(this[p.settings].registrationService)
+          .then(boxes => {
+            if (!Array.isArray(boxes)) {
+              return resolve();
```
azasypkin: __question:__ should we schedule a retry in this case?

app/js/lib/foxbox/foxbox.js
azasypkin: __note:__ not required for this PR, but just wanted to note that I'd rather simplify this method (remove this nasty outer ```new Promise``` and no need in ```timeout``` argument) and reduce the scope for settimeout:
```
_initDiscovery() {
    // For development purposes if you want to skip the
    // discovery phase set the 'foxbox-skipDiscovery' variable to
    // 'true'.
    if (this[p.settings].skipDiscovery) {
      return Promise.resolve();
    }

    return this[p.net].fetchJSON(this[p.settings].registrationService)
      .then(boxes => {
        boxes = Array.isArray(boxes) ? boxes : [];

        ......

        // If the registration server didn't give us any info and
        // a retry.
        if (!this[p.boxes].length &&
          !this[p.settings]._localOrigin &&
          !this[p.settings]._remoteOrigin) {
          console.warn('No boxes found. Retrying...');
          
          return new Promise((resolve) => {
            setTimeout(() => {
              this._initDiscovery().then(resolve, resolve);
            }, 1000);
          });
        }

        ....

        this._dispatchEvent('box-online', true);
      })
      .catch(() => {
        ....
      });
  }
```

app/js/lib/foxbox/foxbox.js
```diff
@@ -123,49 +127,66 @@ export default class Foxbox extends Service {
...
+                !this[p.settings]._localOrigin &&
+                !this[p.settings]._remoteOrigin) {
+              console.warn('No boxes found. Retrying...');
+              return this._initDiscovery(1000);
+            }
+
+            if (!this[p.settings].configured) {
+              this.selectBox();
+            }
+            console.log('Box found');
```
azasypkin: __nit:__ please don't forget to remove ```console.log``` from the final PR (I admit our log is already quite noisy even without this one :)), we'll switch to ```debug``` later). But if you think it's still useful let's make it more informative with something like ```console.log('Box found %s', ..url...of...the...box);``` (if possible).

app/js/lib/foxbox/settings.js
```diff
@@ -89,8 +87,10 @@ export default class Settings extends Model {
 
   set localOrigin(origin) {
     this._localOrigin = origin ?
-        (new URL(origin)).origin : DEFAULT_LOCAL_ORIGIN;
-    storage.setItem(`${PREFIX}localOrigin`, this._localOrigin);
+        (new URL(origin)).origin : undefined;
```
azasypkin: __optional nit:__ can we use ```null``` instead of ```undefined``` - it's initialized with the ```null``` in the constructor (in case we don't have anything in the localStorage it should return ```null```) - just for the consistency.

azasypkin: __nit:__ and it likely can fit into single line now :)

app/js/lib/foxbox/settings.js
azasypkin: I thought that @gmarty was asking to call ```removeItem``` on the localStorage in case ```!this._localOrigin``` to keep these values in sync, but I'm not sure too :)

app/js/views/user-login.jsx
```diff
@@ -8,10 +8,21 @@ export default class UserLogin extends BaseView {
...
     this.state = {
       boxes: props.foxbox.boxes,
-      value: null
+      value: null,
+      loginEnabled: false
     };
 
     this.foxbox = props.foxbox;
+
+    this.handleOnBoxOnline = this.handleOnBoxOnline.bind(this);
```
azasypkin: __optional nit:__ I'd call it just ```onBoxOnline```, but up to you and @gmarty.

azasypkin:
Looks good to me, I've left comment about one potential typo (```_remoteOrigin```), few nits and small optional simplification proposal, r+.

Thanks for improving this!

app/js/lib/foxbox/foxbox.js
gmarty: So what value do we get from the registration server when there's no tunnel set up?

app/js/views/user-login.jsx
gmarty: I think we agreed to use `on...` formatted name for handlers in JSX.

app/js/lib/foxbox/README.md
```diff
@@ -52,3 +52,7 @@ is disconnected from it.
 ### `service-state-change`
 
 This event is emitted when a service states change.
+
+### `box-online`
+
+This event is emitted once a box is discovered on the local network or every time we get a response from a ping to a previously discovered box url (local or remote).
```
gmarty: Can you please insert a break line every 80 characters?

app/js/lib/foxbox/foxbox.js
gmarty: Ah yeah I hadn't realised it was not in the same promise cascade. Looks good to me!

app/js/lib/foxbox/foxbox.js
gmarty: I agree with @azasypkin but that'd be nice to land in this patch so that we don't pollute too much git history and blame.

app/js/lib/foxbox/foxbox.js
gmarty: In my understanding, in this case, we get a response from the registration server but the response is bad (likely a bug in the registration server), so there's probably no need to try again.
@ferjm: What do you think?

app/js/lib/foxbox/settings.js
gmarty: Yes sorry, my wording was really awkward. @azasypkin got it right, do we need to call `removeItem` if the value is falsey?

app/js/lib/foxbox/foxbox.js
ferjm: We get null

app/js/lib/foxbox/settings.js
ferjm: Oh, ok. Yes, we can do that. Although I think we shouldn't get to that point.

app/js/lib/foxbox/foxbox.js
ferjm: Yes, I think in this case it's pointless to retry. We should probably just warn and bail out.

azasypkin: Ok, yep, makes sense

PR #97, opened 2016-04-15
Add Door Lock support. r=gmarty

azasypkin:


app/js/views/services-list-item.jsx
```diff
@@ -222,11 +222,21 @@ export default class ServicesListItem extends React.Component {
       case 'ip-camera@link.mozilla.org':
         return this.renderGenericService('Camera', 'ip-camera');
       case 'OpenZwave Adapter':
+        if (this.hasDoorLockChannel(this.props.getters) ||
```
azasypkin: There is no other way to distinguish them for now :/

gmarty:
That looks good to me! r+

azasypkin:
> That looks good to me! r+

Thanks! Landing then, fabrice also managed to check it on real lock - he didn't report any issue so far.

JohanLorenzo:
I checked with the lock in Paris. I managed to create recipes that lock or unlock the door. One piece of trivia, you can create a rule that locks the door when foxbox detects it's unlocked.

azasypkin:
> I checked with the lock in Paris. I managed to create recipes that lock or unlock the door. It correctly detected even if the Z-wave dongle deals with another sensor (like the motion one)

Yay, thanks for verifying! Then clearly I have issue with the emulator - will investigate.

PR #100, opened 2016-04-18
In recipes, reset subsequent values when a field is updated. r=azasypkin

gmarty:


azasypkin:
Looks good, thanks! r+

PR #106, opened 2016-04-19
Do not ping to <origin>/null/ping if reg server ping fails. r=azasypkin

ferjm:
r? @azasypkin

### DIFF
--- a/app/js/lib/foxbox/foxbox.js
+++ b/app/js/lib/foxbox/foxbox.js
@@ -175,12 +175,7 @@ export default class Foxbox extends Service {
         if (!this[p.boxes].length &&
             !this[p.settings]._localOrigin &&
             !this[p.settings]._tunnelOrigin) {
-          console.warn('No boxes found. Retrying...');
-          return new Promise((resolve) => {
-            setTimeout(() => {
-              this._initDiscovery().then(resolve, resolve);
-            }, 1000);
-          });
+          throw new Error('Registration service did not return any boxes.');
         }
 
         if (!this[p.settings].configured) {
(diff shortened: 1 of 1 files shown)


app/js/lib/foxbox/foxbox.js
```diff
@@ -190,10 +190,19 @@ export default class Foxbox extends Service {
...
       .catch(() => {
-        // Default to a previously stored box registration.
         if (this[p.settings]._localOrigin ||
             this[p.settings]._tunnelOrigin) {
+          // Default to a previously stored box registration.
           this._dispatchEvent('box-online', true);
+        } else {
+          // If there's no previosly stored box registration, we schedule a
+          // retry.
+          console.warn('No boxes found. Retrying...');
```
azasypkin: __question:__ since we have the same code in two places now, @ferjm how do you feel about changing the code above to something like this?
```
// If the registration server didn't give us any info and
// a retry.
if (!this[p.boxes].length &&
    !this[p.settings]._localOrigin &&
    !this[p.settings]._tunnelOrigin) {
  throw new Error('Registration service did not return any boxes.');
}
```

So now our ```catch``` clause is the only place where we make a retry. + Add error we get to the console warning. 
```
console.warn('Retrying box discovery... Reason is %o', e);
```

P.S. Error messages are just drafts - I believe you know better what error message should tell exactly :)

app/js/lib/foxbox/foxbox.js
```diff
@@ -190,10 +190,19 @@ export default class Foxbox extends Service {
         this._dispatchEvent('box-online', true);
       })
       .catch(() => {
-        // Default to a previously stored box registration.
         if (this[p.settings]._localOrigin ||
             this[p.settings]._tunnelOrigin) {
+          // Default to a previously stored box registration.
           this._dispatchEvent('box-online', true);
+        } else {
+          // If there's no previosly stored box registration, we schedule a
```
azasypkin: __typo:__ ```previously```

azasypkin:
Looks good to me, just one tiny typo and small proposal (in case you think it's worth it). In any case it's r+.

Thanks!

PR #109, opened 2016-04-20
The recipe must be initialised in the critical path. r=azasypkin

gmarty:
If not, landing on a the Recipes page when refreshing the app won't work.

### DIFF
--- a/app/js/lib/foxbox/foxbox.js
+++ b/app/js/lib/foxbox/foxbox.js
@@ -76,30 +76,31 @@ export default class Foxbox extends Service {
   init() {
     window.foxbox = this;
 
-    // No need to block the UI on the discovery process.
-    // Once we discover a box we can connect to, we will start
-    // polling and triggering box-online events with a boolean
-    // indicating if we have access to box or not.
-    this._initDiscovery().then(() => {
-      return this[p.net].init();
-    })
-    .then(() => {
-      // Start polling.
-      this[p.settings].on('pollingEnabled', () => {
-        this.togglePolling(this[p.settings].pollingEnabled);
-      });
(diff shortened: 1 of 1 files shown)


app/js/lib/foxbox/foxbox.js
```diff
@@ -89,15 +89,15 @@ export default class Foxbox extends Service {
...
-        settings: this[p.settings],
-        net: this[p.net]
-      });
     });
 
     return this._initUserSession()
       .then(() => {
+        this.recipes = new Recipes({
+          settings: this[p.settings],
+          net: this[p.net]
```
azasypkin: Looks good, the only thing that worries me a bit is that ```net``` is initialized once discovery is resolved.

azasypkin:
r+, just one concern, but maybe it's okay for now.

gmarty:
I reorganised the sequence to initialise the network first then the rest as of today.

azasypkin:
Hmm, looks like I'm facing similar thing in https://github.com/fxbox/app/issues/116.

This "chicken and the egg problem" blocking me right now, but I have one idea in mind that I'm trying to implement.

azasypkin:
> but I have one idea in mind that I'm trying to implement.

Okay it works well with DB, but not with Network :/ For DB idea is to keep internal initialization promise that is used by all async operations, so you can call any async function on DB and it will be performed and resolved only when initialization promise is resolved - so it safe to call such methods at any time. But Network is different. Wondering if should just make Network simple and make Foxbox to manage local/tunnel origin...

gmarty:
We can close this one now that https://github.com/fxbox/app/pull/120 landed.

PR #110, opened 2016-04-20
Re-added support for Motion Sensor and DoorLock (using separate classes). r=gmarty

azasypkin:


app/js/lib/foxbox/services/door-lock.js
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+   * Checks if the lock is locked.
+   *
+   * @returns {Promise.<boolean>}
+   */
+  isLocked() {
+    return this.get('DoorLocked')
+      .then((response) => {
+        if (!response) {
```
azasypkin: For OpenZWave we can have ```nulll``` response that means status is not available yet.

app/js/lib/foxbox/services/light.js
```diff
@@ -15,13 +15,11 @@ export default class LightService extends BaseService {
   }
 
   isAvailable() {
-    return this.get('available')
-      .then((response) => response[Object.keys(response)[0]].OnOff === 'On');
+    return this.get('available').then((response) => response.OnOff === 'On');
```
azasypkin: Would be great if you can check that on real device for me :)

app/js/lib/foxbox/services/ip-camera.js
```diff
@@ -14,10 +14,6 @@ export default class IpCameraService extends BaseService {
     return TYPE;
   }
 
-  getImageList() {
-    return this.get('image_list');
-  }
```
gmarty: Why removing this? Because it's not used or because the adapter changed in the box?

azasypkin: Like we did with ```turnOn``` and ```turnOff``` for the light service - because it's unused - so less code to maintain until we really need it. But tell me if you want me to return it back.

app/js/lib/foxbox/services/door-lock.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /**
+   * Toggles state of the lock depending on the "isLocked" parameter.
+   *
+   * @param isLocked Boolean value indicating whether we want lock or unlock
+   * the lock.
+   * @returns {Promise}
+   */
+  toggle(isLocked) {
```
gmarty: I find `toggle()` confusing because the common case for a function called `toggle` is to reverse the state without passing parameters (e.g. `el.classList.toggle('highlighted');`).
But even `lock()` sounds confusing.

app/js/lib/foxbox/services/motion-sensor.js
```diff
@@ -0,0 +1,16 @@
...
+
+export default class MotionSensorService extends BaseService {
+  constructor(props, config) {
+    super(props, config);
+    Object.seal(this);
+  }
+
+  get type() {
+    return TYPE;
+  }
```
gmarty: We need basic setter/getter here even if we don't use them right now, we'll need them in the future.

app/js/lib/foxbox/services/door-lock.js
azasypkin: > I find toggle() confusing because the common case for a function called toggle is to reverse the state without passing parameters (e.g. el.classList.toggle('highlighted');).

yeah, it's more like toggle with ```force``` parameter, ```el.classList.toggle('highlighted', true/false)```.

I can either have lock/unlock pair or toggle or lockUnlock, what do you prefer?

app/js/lib/foxbox/services/ip-camera.js
gmarty: `turnOn` and `turnOff` were syntactic sugar on top of the setters. I think we need to add methods for all the getters and setters of the adapter, so that we don't have to debug the box responses.
That'll be valuable too when we generate documentation from the code.

app/js/lib/foxbox/services/motion-sensor.js
azasypkin: Mm, didn't get, what you want me to add?

app/js/lib/foxbox/services/door-lock.js
gmarty: `lockUnlock` is verbose but I think that's the less ambiguous one. What do you think?

app/js/lib/foxbox/services/ip-camera.js
azasypkin: Right, but I believe API will definitely change so didn't want to do double job constantly updating method that nobody uses :)

But not a big deal, tell and I return it back :)

app/js/lib/foxbox/services/door-lock.js
azasypkin: I'm fine with ```lockUnlock```

app/js/lib/foxbox/services/motion-sensor.js
gmarty: Something like `isPresenceDetected` (or `detect`).

azasypkin: Well.... I intentionally didn't add it, it's somewhat special thing, that will automatically change in 30 seconds (configurable though). My rationale was - until we get reliable method of keeping this properties in sync with foxbox there is a little value in them, they will be just confusing...

But if you mean method not getter/setter property, I can add but still won't use it until we get polling/websocket support - otherwise people will complain for half-working feature :)

gmarty: OK, let's not implement it now. We'll look into it when we actually need it.

app/js/lib/foxbox/services/light.js
gmarty: Works like a charm! And is much much nicer.

app/js/lib/foxbox/services/motion-sensor.js
azasypkin: Yep! (that's not my laziness, I just feel API will change again, that's why I'm trying to implement only something that is needed :))

app/js/lib/foxbox/services/ip-camera.js
gmarty: Not sure what to do then. If we never use it in the app, this getter will never be implemented in this class. When one day this lib is independent and used by 3rd party dev, it'll be confusing that not all getters or setters are exposed.

But if the API is likely to change, let's remove it for now.

azasypkin: I totally see your point and agree with you, I'm just waiting for more stable API state. We just did several things and didn't have time to bulletproof API in full, IMO there is high probability that some big pieces can change. But once we know that API reaches some milestone - then no doubt we'll need to expose methods that we don't use, but others can.

azasypkin: + it's a good first bug for any contributor that would like to work on the app :)

app/js/views/services-list-item.jsx
```diff
@@ -28,6 +29,15 @@ export default class ServicesListItem extends React.Component {
           })
           .catch(console.error.bind(console));
         break;
+      case 'door-lock':
+        this.service.isLocked()
+          .then((locked) => {
+            this.setState({ locked });
+          })
+          .catch((e) => {
```
gmarty: nit: Let's avoid 1 letter variable names and use more meaningful names instead, like `error` or `err`.

app/js/views/services-list-item.jsx
```diff
@@ -45,6 +55,19 @@ export default class ServicesListItem extends React.Component {
       });
   }
 
+  onDoorLockUnlock(e) {
+    const locked = e.target.checked;
+
+    this.setState({ locked });
+
+    this.service.toggle(locked)
+      .catch((e) => {
```
gmarty: nit: same here.

app/js/views/services-list-item.jsx
```diff
@@ -45,6 +55,19 @@ export default class ServicesListItem extends React.Component {
       });
   }
 
+  onDoorLockUnlock(e) {
```
gmarty: nit: same here (`evt` or `event`).

gmarty:
I left a few nits but r+!

app/js/views/services-list-item.jsx
azasypkin: Yep, makes sense

PR #112, opened 2016-04-21
Enforce more coding styles with eslint. r=@azasypkin

gmarty:
This pull request makes eslint to enforce more coding style. The following new rules are set:
## [Possible errors](http://eslint.org/docs/rules/#possible-errors)
* valid-jsdoc

## [Best practices](http://eslint.org/docs/rules/#best-practices)
* eqeqeq
* no-case-declarations
* no-else-return
* no-extra-bind
* no-implicit-coercion
* no-multi-spaces
* no-throw-literal
* no-warning-comments

## [es6](http://eslint.org/docs/rules/#ecmascript-6)
* arrow-body-style
* arrow-spacing
* object-shorthand
* prefer-const
* prefer-template

No [stylistic rules](http://eslint.org/docs/rules/#stylistic-issues) have been added in this patch. They'll be considered in the future.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,12 +22,43 @@
     "no-undef": ["error"],
     "no-unused-vars": ["error", { "vars": "all", "args": "after-used" }],
     "no-use-before-define": ["error", { "functions": false, "classes": true }],
-    "no-var": ["error"],
     "quotes": ["error", "single"],
     "semi": ["error", "always"],
     "comma-dangle": ["error", "always-multiline"],
+    "guard-for-in": ["error"],
+    "valid-jsdoc": ["error",
+      {
+        "prefer": {
+          "return": "return"
+        },
+        "preferType": {
+          "String": "string",
+          "Number": "number",
+          "Boolean": "boolean",
+          "object": "Object",
+          "array": "Array"
+        },
+        "requireReturn": false,
+        "requireParamDescription": false,
+        "requireReturnDescription": false
+      }
+    ],
+    "eqeqeq": ["error"],
+    "no-case-declarations": ["error"],
+    "no-else-return": ["error"],
+    "no-extra-bind": ["error"],
+    "no-implicit-coercion": ["error", { "boolean": false }],
+    "no-multi-spaces": ["error"],
+    "no-throw-literal": ["error"],
+    "no-warning-comments": ["error", { "terms": ["todo", "@todo:", "fixme",
+      "xxx"] }],
+    "arrow-body-style": ["warn", "as-needed"],
     "arrow-parens": ["error", "always"],
-    "guard-for-in": ["error"]
+    "arrow-spacing": ["error"],
+    "no-var": ["error"],
+    "object-shorthand": ["error"],
+    "prefer-const": ["warn"],
+    "prefer-template": ["error"]
   },
 
   "env": {
--- a/app/js/lib/foxbox/recipes.js
+++ b/app/js/lib/foxbox/recipes.js
@@ -54,8 +54,8 @@ export default class Recipes {
       'POST',
       { getters: [{ kind: 'ThinkerbellRuleSource' }] }
     )
+    // Mark getters and setters with more friendly names.
     .then((services) => {
-      // Mark getters and setters with more friendly names.
       return services.map((service) => {
         const enabledGetterId = Object.keys(service.getters).find(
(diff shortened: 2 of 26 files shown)


.eslintrc
```diff
@@ -22,12 +22,43 @@
...
+        "preferType": {
+          "String": "string",
+          "Number": "number",
+          "Boolean": "boolean",
+          "object": "Object",
+          "array": "Array"
+        },
+        "requireReturn": false,
+        "requireParamDescription": false,
+        "requireReturnDescription": false
```
gmarty: The JSDoc annotations checking is quite permissive at the moment. We'll make it stricter when we generate documentation from the code.

azasypkin: Yep, makes sense

.eslintrc
```diff
@@ -22,12 +22,43 @@
...
+      }
+    ],
+    "eqeqeq": ["error"],
+    "no-case-declarations": ["error"],
+    "no-else-return": ["error"],
```
azasypkin: \0/ -ahh my favorite ```no-else-return```! :)

app/js/lib/foxbox/recipes.js
```diff
@@ -54,9 +54,8 @@ export default class Recipes {
       'POST',
       { getters: [{ kind: 'ThinkerbellRuleSource' }] }
     )
-    .then((services) => {
       // Mark getters and setters with more friendly names.
-      return services.map((service) => {
+      .then((services) => services.map((service) => {
```
azasypkin: __question:__ When closing bracket is moved to a separate line above it looks nicer when ```.then``` is aligned with it, what do you think?

app/js/lib/foxbox/webpush.js
```diff
@@ -80,10 +80,10 @@ export default class WebPush {
       })
       .catch((error) => {
         if (Notification.permission === 'denied') {
-          throw 'Permission request was denied.';
+          throw new Error('Permission request was denied.');
         } else {
```
azasypkin: __nit:__ Ha! I expected ESLint to complain about ```else``` if we clearly have only throw in the ```if``` section :)

tests/common/login_test.js
```diff
@@ -11,26 +11,16 @@ describe('Login', function() {
     app = new App();
   });
 
-  beforeEach(() => {
-    return app.init()
-      .then((defaultView) => { loginView = defaultView; });
-  });
+  beforeEach(() => app.init()
```
azasypkin: Honestly here it became even less readable than it was before :)

azasypkin:
Looks really good! The only thing I don't really like is this:
```
(list) => list.map((item) => {
  ......
  ......
  return item;
});
```

I see ```return item``` and think that top function returns ```item``` but in fact it returns list of items.... Maybe it's just me, so up to you.

r+ with nits handled, rebased on master and green travis :)

app/js/lib/foxbox/recipes.js
gmarty: About question 2: Yes, this style is enforced by `arrow-body-style`. I think you're right, it doesn't always make this easier to read. I suggest to make it only a warning for now so that we can decide whether to use it or not. I'll also revert the ones that are hard to read. Is that OK?

azasypkin: Yep, sounds good! (I think eventually we'll be able to make it ```error``` again, since big ```map``` callbacks almost asking for a dedicated function anyway :))

app/js/lib/foxbox/webpush.js
gmarty: I see, you think it should work like the `no-else-return`. I don't think there is a specific rule for that. So maybe we should open an issue on eslint or write and contribute our own rule.

tests/common/login_test.js
gmarty: See note above.

app/js/lib/foxbox/webpush.js
azasypkin: Yep, I was thinking about ```no-else-return```...

gmarty:
I reverted the `arrow-body-style` for the call that were not final (like return a `obj.map()` or `obj.then()`) and made this setting optional so we can decide whether to use the syntax or not.

gmarty:
I had to relax the `prefer-const` option because Firefox has issue with const in for loops. Hopefully Travis is green this time.

PR #115, opened 2016-04-22
Import Event Dispatcher and tests. r=gmarty

azasypkin:


app/js/lib/foxbox/event-dispatcher.js
```diff
@@ -206,42 +196,25 @@ export default class EventDispatcher {
...
-  if (!target || typeof target !== 'object') {
-    throw new Error('Object to mix into should be valid object!');
-  }
+  /**
+   * Checks if there are any listeners that listen for the specified event.
+   *
+   * @param {string} eventName Name of the event to check listeners for.
+   * @returns {boolean}
+   */
+  hasListeners(eventName) {
```
azasypkin: I have mixed feelings about this method - but it simplifies a lot my next PR, so I've added this one. If in the future I find better approach to know that I can shutdown the bus (or something that depends on the bus) if there is no listeners anymore, I would probably try to get rid of this method.

azasypkin:
@gmarty r?

app/js/lib/foxbox/event-dispatcher.js
```diff
@@ -169,7 +157,8 @@ export default class EventDispatcher {
 
   /**
    * Removes all registered listeners for the specified event.
-   * @param {string} eventName Name of the event to remove all listeners for.
+   *
+   * @param {string} [eventName] Name of the event to remove all listeners for.
```
gmarty: nit: Do the brackets around the param name for optional? If we use Closure Compiler flavour of JSDoc, you can indicate optional params with a `=` (see https://developers.google.com/closure/compiler/docs/js-for-compiler#types):
```js
/**
 * Removes all registered listeners for the specified event.
 *
 * @param {string=} eventName Name of the event to remove all listeners for.
 */
```

app/js/lib/foxbox/event-dispatcher.js
```diff
@@ -193,8 +182,9 @@ export default class EventDispatcher {
   /**
    * Emits specified event so that all registered handlers will be called
    * with the specified parameters.
+   *
    * @param {string} eventName Name of the event to call handlers for.
-   * @param {Object} parameters Optional parameters that will be passed to
+   * @param {Object} [parameters] Optional parameters that will be passed to
```
gmarty: nit: see above.

app/js/lib/foxbox/event-dispatcher.js
azasypkin: Btw, does webstorm + enabled ESLint recognize ```{string=}``` as optional parameter for you?

gmarty:
Note on `event-dispatcher.js`: this file looks good. I would have called the `ensure*()` functions something like `assert*()`. But apart from a couple of nits, I'm happy with it.
Reviewing the rest of the PR now.

app/js/lib/foxbox/event-dispatcher.js
gmarty: It works for me on WebStorm but I don't think I set up ESLint properly. If it's a bug maybe we should fill a ticket in WebStorm or ESLint.

tests/unit/lib/foxbox/event-dispatcher_test.js
```diff
@@ -0,0 +1,423 @@
+import EventDispatcher from 'js/lib/foxbox/event-dispatcher';
+
+describe('EventDispatcher >', function () {
```
gmarty: question: Do we have a good reason to use `function () {` as opposed to `() => {` in the test files?

tests/unit/lib/foxbox/event-dispatcher_test.js
```diff
@@ -0,0 +1,423 @@
...
+
+        restrictedEventTarget.emit('allowed-event-1');
+
+        sinon.assert.notCalled(unexpectedHandler);
+        sinon.assert.calledOnce(expectedHandler1);
+        sinon.assert.calledOnce(expectedHandler2);
+        sinon.assert.callOrder(expectedHandler1, expectedHandler2);
+      });
+    });
+  });
```
gmarty: The new `hasListeners()` method needs to be tested too.

gmarty:
r+ when you add a test case for `hasListeners`.
I like that we get all those tests for free!

tests/unit/lib/foxbox/event-dispatcher_test.js
azasypkin: Probably not, "historically" I used ```function``` in tests and where I don't need to grab context (likely browser knows that and doesn't do anything especial, but I don't know for sure), so just used to do it.

Do you think we should use arrows here?

gmarty: It's just an idea.
That would result in a more readable file, but that's debatable.

app/js/lib/foxbox/event-dispatcher.js
azasypkin: Yeah, likely a bug somewhere, will file one if not forget :)

tests/unit/lib/foxbox/event-dispatcher_test.js
azasypkin: Okay, I'll keep like it's currently, but we can discuss this later, I don't have a preference actually

azasypkin:
> I would have called the ensure*() functions something like assert*()

Yeah, maybe ```assert*```-like name is better, I'll rename!

azasypkin:
> r+ when you add a test case for hasListeners.

Ha-ha, I knew you'll notice that :) Will add few tests for this new method as well.

> I like that we get all those tests for free!

Almost for free :p It required some routine ```describe\it\before\after\const\let```-fication :)

azasypkin:
Nits fixed, travis is green - merging!

Issue #107, opened 2016-04-20
Start using EventDispatcher for Foxbox and Services


azasypkin:
Preliminary we want to import [1] and [2] including all existing unit tests. This will require some changes to make it more "es6-classy". And we need to decided whether we want it to be base class or mixin, or maybe both.

[1] https://github.com/mozilla-b2g/gaia/blob/master/shared/js/event_dispatcher.js
[2] https://github.com/vargin/event-dispatcher/blob/master/event-dispatcher.es6.js

azasypkin:
It's being covered in https://github.com/fxbox/app/issues/111 - MotionSensorService will be the first service that uses custom events, the rest of the services will expose events as needed.

Issue #99, opened 2016-04-18
Position/order of the known services in Home view should be preserved


azasypkin:
Currently if we have let's say Service 1, Service 2 and Service 3. They are saved in Indexed DB, when user opens the app we display services in particular order, but once we receive service list from the box the order of these services can change sometimes even if we still have the same services.

IIRC @fabricedesre was the one who was particularly complaining about this weird behavior :)

azasypkin:
We've fixed the most annoying/buggy part of this issue - refreshing of the service order without any reason :)

So adjusting scope here - we want to introduce some deterministic order based on sorting/grouping by service type and name.

PR #120, opened 2016-04-22
 Introduce Foxbox API sub-module. Maintain proper services in-memory cache in a separate submodule. Use EventDispatcher for Network, Settings and WebPush sub-modules. (fixes #116). r=gmarty

### LINKED ISSUE #116: We should remove services that no longer available from IndexedDB
It becomes a __huge__ problem if IndexedDB contains something that is no longer available in this case we always think that number of services has changed and we fire corresponding events and trying to re-render UI and more...

azasypkin:


app/js/lib/foxbox/api.js
```diff
@@ -0,0 +1,41 @@
+const p = Object.freeze({
+  settings: Symbol('settings'),
+  net: Symbol('net'),
+
+  // Private methods.
+  getURL: Symbol('getURL'),
+});
+
+export default class API {
```
azasypkin: One of the ideas I'm still forming in my head is to make API work a bit like DB - it will be initialized immediately, but since all its methods are sync they will wait until user is authenticated and box is discovered....

app/js/lib/foxbox/network.js
```diff
@@ -177,29 +179,30 @@ export default class Network {
...
+      .then(([isLocalOnline, isRemoteOnline]) => {
+        const currentState = isLocalOnline || isRemoteOnline;
+
+        this[p.local] = isLocalOnline;
+        this[p.remote] = isRemoteOnline;
+
+        if (currentState !== previousState) {
+          this.emit('online', currentState);
+        }
+      });
```
gmarty: What if one promise hangs on for a long time? It will delay the other one. I still think it's better not to use `Promise.all()` here.

app/js/lib/foxbox/webpush.js
```diff
@@ -1,37 +1,36 @@
...
   listenForMessages: Symbol('listenForMessages'),
-  dispatchEvent: Symbol('dispatchEvent'),
 });
 
-export default class WebPush {
-  constructor(network, settings, callback) {
-    this[p.net] = network;
+export default class WebPush extends EventDispatcher {
+  constructor(api, settings) {
+    super(['push-action']);
```
gmarty: nit: we need to have all the event names on the same model. I like the preterite verb model like `events-changed` or`event-state-changed`.

azasypkin: Yep, we need to agree on the event namings for sure. Lower case, dashes to separate words and "preterite verb model" - sounds good?

Here we can name it as ```push-message-received``` or ```push-action-received```.

app/js/lib/foxbox/services.js
```diff
@@ -0,0 +1,246 @@
...
+
+    if (typeA !== 'object' || valueA === null || valueB === null) {
+      return valueA !== valueB;
+    }
+
+    return !isSimilar(valueA, valueB);
+  });
+};
+
+export default class Services extends EventDispatcher {
```
gmarty: I'm not a big fan of changing the signature of the `foxbox` instance and use `foxbox.service` for everything related to the services, Splitting the logic in multiple files makes sense for the code organisation but that shouldn't impact the publicly exposed API. It seems longer and unnecessarily complicated to me. Or maybe I just need to get used to it.
What do you think?
What else is exposed from the `foxbox` instance? (e.g. `foxbox.boxes`, `foxbox.clear()`...)?

app/js/lib/foxbox/api.js
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Instance of the API class is intended to abstract consumer from the API
+ * specific details (e.g. API base URL and version). It also tracks
+ * availability of the network, API host and whether correct user session is
+ * established. If any of this conditions is not met all API requests are
+ * blocked until it's possible to perform them, so consumer doesn't have to
+ * care about these additional checks.
+ */
+export default class API {
```
gmarty: :+1: 
I like this new API class. It makes the code easier to read.

app/js/lib/foxbox/webpush.js
gmarty: Yeah that sounds good to me.

gmarty:
I left a few comments.
There's a lot in this PR, so the commit name should be updated to account for the bigger scope. It looks good to me.

app/js/lib/foxbox/services.js
azasypkin: > What do you think?

Well, for me ```foxbox.services``` looks almost the same as ```foxbox.recipes```, so I feel good with ```foxbox.services``` :) But if you're strongly against this, I'm totally fine with making it similar to what it was before.

> What else is exposed from the foxbox instance? (e.g. foxbox.boxes, foxbox.clear()...)?

Yeah, it's box management (```.boxes```, ```selectBox```), user session management (```login```, ```logout```), notification management and aggregation of box submodules (```.recipes``` and now ```.services```).

So to summarize, at this stage I'm OK with any choice, just let me know what you prefer.

app/js/lib/foxbox/network.js
azasypkin: Yeah, this is a valid concern! Let me think what to do here.

app/js/lib/foxbox/services.js
gmarty: Ah yeah I forgot about `foxbox.recipes`. I need to give it some thoughts, don't change anything for now. This approach is probably the best one.

tests/unit/lib/foxbox/services_test.js
```diff
@@ -0,0 +1,195 @@
+import { waitFor } from '../../test-utils';
```
azasypkin: Here not all cases covered, just most important and obvious. Should be good enough to start with.

azasypkin:
Should also fix https://github.com/fxbox/app/pull/109, https://github.com/fxbox/app/issues/107 and https://github.com/fxbox/app/issues/99

azasypkin:
I think it should be ready for review, r+ @gmarty

app/js/lib/foxbox/foxbox.js
```diff
@@ -23,55 +19,33 @@ const p = Object.freeze({
...
     super();
 
     // Private properties.
     this[p.settings] = new Settings();
     this[p.db] = new Db();
-    this[p.net] = new Network(this[p.settings], (foxboxOnline) => {
-      this._dispatchEvent('box-online', foxboxOnline);
+
+    this[p.net] = new Network(this[p.settings]);
+    this[p.net].on('online-state-changed', (online) => {
```
azasypkin: Naming of events is hard :/ Tell me if you have better names for the events!

app/js/lib/foxbox/foxbox.js
```diff
@@ -87,15 +61,10 @@ export default class Foxbox extends Service {
...
         // Start polling.
-        this[p.settings].on('pollingEnabled', () => {
-          this.togglePolling(this[p.settings].pollingEnabled);
-        });
-        this.togglePolling(this[p.settings].pollingEnabled);
-
-        this.recipes = new Recipes({
-          settings: this[p.settings],
-          net: this[p.net],
+        this[p.settings].on('polling-setting-changed', () => {
```
azasypkin: I left it here as it was to reduce affected surface for this PR is already huge :) But in the future I was thinking that services itself can trigger that...

app/js/lib/foxbox/foxbox.js
```diff
@@ -180,32 +149,29 @@ export default class Foxbox extends Service {
         // we have no record of previous registrations, we schedule
         // a retry.
         if (!this[p.boxes].length &&
-            !this[p.settings]._localOrigin &&
```
azasypkin: I'm not sure why private fields were used here.

app/js/lib/foxbox/foxbox.js
```diff
@@ -180,32 +149,29 @@ export default class Foxbox extends Service {
...
+            !this[p.settings].localOrigin &&
+            !this[p.settings].tunnelOrigin) {
           throw new Error('Registration service did not return any boxes.');
         }
 
         if (!this[p.settings].configured) {
           this.selectBox();
         }
-
-        this._dispatchEvent('box-online', true);
```
azasypkin: Actually box is not online until we receive any ping response from local or remote box. It's just registration service online, so I've removed that.

app/js/lib/foxbox/foxbox.js
```diff
@@ -180,32 +149,29 @@ export default class Foxbox extends Service {
...
       })
       .catch((error) => {
-        if (this[p.settings]._localOrigin ||
-            this[p.settings]._tunnelOrigin) {
+        if (this[p.settings].localOrigin || this[p.settings].tunnelOrigin) {
           // Default to a previously stored box registration.
-          this._dispatchEvent('box-online', true);
```
azasypkin: Same here

app/js/lib/foxbox/network.js
```diff
@@ -172,49 +175,70 @@ export default class Network {
...
 
   /**
-   * Get a URL but don't process the response.
+   * Performs HTTP 'GET' request to the specified URL. Returns 'true' if
+   * response was successful (any of 200-299 status codes) or 'false'
+   * otherwise.
    *
    * @param {string} url The URL to send the request to.
-   * @return {Promise}
+   * @return {Promise<boolean>}
```
azasypkin: Looks like ```boolean``` is what we need from this method (and logged error if any) right now. If in the future we have rich ping response we can reconsider this.

app/js/lib/foxbox/network.js
```diff
@@ -172,49 +175,70 @@ export default class Network {
...
+   * 'online-state-changed' event.
+   *
+   * @param {boolean} previousOnlineState Previous 'online' state.
+   * @param {Symbol} localOrRemote Symbol indicating whether we process local
+   * pong or remote one.
+   * @param {boolean} isOnline Flag that indicates whether pinged server
+   * successfully responded to ping request.
+   * @private
+   */
+  [p.onPong](previousOnlineState, localOrRemote, isOnline) {
```
azasypkin: I don't really happy about this method, alternative was to create function inside ```pingBox``` on every ping. And I don't like alternative even more...

app/js/lib/foxbox/services.js
```diff
@@ -0,0 +1,273 @@
...
+          cache.set(service.id, service);
+
+          if (isExistingService) {
+            this.emit('service-changed', service);
+          } else {
+            servicesToAddCount++;
+          }
+        });
+
+        const servicesToRemoveCount = storedServices.length +
```
azasypkin: A bit nasty calculation, but it allows us to avoid unnecessary iterations over stored services, so we only do it when we really need to remove anything.

app/js/lib/foxbox/settings.js
```diff
@@ -29,29 +29,26 @@ const storage = localStorage ? localStorage : {
   clear: () => {},
 };
 
-export default class Settings extends Model {
+export default class Settings extends EventDispatcher {
   constructor() {
+    super(['session-changed', 'polling-setting-changed']);
```
azasypkin: Honestly it feels like session should not belong to ```Settings```, but that's what we have now, and I'm trying to differentiate settings from non-settings with ```-setting-``` part in the event name....

app/js/lib/foxbox/settings.js
```diff
@@ -121,13 +111,19 @@ export default class Settings extends Model {
...
+    if (this._session === session) {
+      return;
+    }
+
+    this._session = session;
+
+    if (this._session) {
       storage.setItem(`${PREFIX}session`, this._session);
+    } else {
+      storage.removeItem(`${PREFIX}session`);
```
azasypkin: as far as I can tell, we should always remove session from localstorage if ```session``` is not non-empty string, that's why I'm changing this. Am I missing something?

app/js/lib/foxbox/settings.js
```diff
@@ -145,9 +141,14 @@ export default class Settings extends Model {
   }
 
   set pollingEnabled(value) {
-    value = !!value;
```
azasypkin: I haven't found any place where we pass something non-boolean here and basically we should not.

app/js/views/tag-item.jsx
```diff
@@ -13,28 +13,7 @@ export default class TagItem extends React.Component {
   }
 
   handleSetTag(evt) {
-    const checked = evt.target.checked;
```
azasypkin: It's not properly supported right now so it just introduces maintenance burden, _but_ supporting tags is the next thing I'm going to do after watchers patch.

tests/unit/test-utils.js
```diff
@@ -0,0 +1,10 @@
+import Rx from 'rxjs';
+
+export function waitFor(fn) {
```
azasypkin: Tell me if you prefer class with static methods instead. I found directly exported functions even more handy, but I don't have strong opinion.

app/js/lib/foxbox/foxbox.js
```diff
@@ -23,55 +19,33 @@ const p = Object.freeze({
...
-        this._dispatchEvent('push-action', msg);
-      });
+    this[p.api] = new API(this[p.net], this[p.settings]);
 
-    // Public properties.
-    this.recipes = null;
+    this[p.webPush] = new WebPush(this[p.api], this[p.settings]);
+    this[p.webPush].on('message-received', (msg) => {
+      this._dispatchEvent('push-message-received', msg);
+    });
```
gmarty: Maybe move the events listeners to the `init()` method instead of the constructor.
This may provide some optimisation where we delay the costly init after a basic view is rendered.
We could also add a `teardown` method where we remove these event listeners.

app/js/lib/foxbox/foxbox.js
gmarty: Sounds good for now.

app/js/lib/foxbox/foxbox.js
azasypkin: > Maybe move the events listeners to the init() method instead of the constructor.

Yeah, haven't thought about this, but yeah I'll move attaching event handlers to ```init```.

app/js/lib/foxbox/foxbox.js
gmarty: That's because the `Settings` class needed updating.

azasypkin: Hmm, I thought we had public ```localOrigin/hostName``` and ```tunnelOrigin/hostName``` properties for some time already, but nevermind :)

app/js/lib/foxbox/services.js
```diff
@@ -0,0 +1,273 @@
...
+        }
+      });
+  }
+
+  /**
+   * Creates specific service instance using raw data returned from server. If
+   * there is no specific service class provided, BaseService class is used.
+   *
+   * @param {Object} data Service raw data.
+   * @return {BaseService}
```
gmarty: It's not always an instance of `BaseService`, even if it always extends it. Not sure if we can instruct JSDoc about that. If not, `{Object}` is the most appropriate.

app/js/lib/foxbox/settings.js
gmarty: That sounds good, yeah.

app/js/lib/foxbox/settings.js
```diff
@@ -29,29 +29,26 @@ const storage = localStorage ? localStorage : {
...
+    this._configured = storage.getItem(`${PREFIX}configured`) !== null ?
+      storage.getItem(`${PREFIX}configured`) === 'true' : false;
+    this._localOrigin = localOrigin;
+    this._tunnelOrigin = storage.getItem(`${PREFIX}tunnelOrigin`) || '';
+    this._clientId = storage.getItem(`${PREFIX}clientId`) || '';
+    this._session = storage.getItem(`${PREFIX}session`);
+    this._skipDiscovery = storage.getItem(`${PREFIX}skipDiscovery`) === 'true';
+    this._pollingEnabled = pollingEnabled;
+    this._pushEndpoint = storage.getItem(`${PREFIX}push_endpoint`) || null;
+    this._pushPubKey = storage.getItem(`${PREFIX}push_pubKey`) || null;
```
gmarty: These all should really be private members exposed via getters and optional setters.
If not in this PR, we should create an issue to keep a track of it.

azasypkin: Sure, I'll file issue for that.

app/js/views/tag-item.jsx
gmarty: Yes. IIRC you were throwing in earlier version of this PR. Maybe a `console.error` would be good to have here too.

tests/foxbox-simulator/http-server.js
```diff
@@ -9,4 +9,8 @@ foxboxSimulator.get('/', (req, res) => {
   res.status(200).sendFile(path.join(`${__dirname}/static/login.html`));
 });
 
+foxboxSimulator.get('/ping', (req, res) => {
+  res.status(200).end();
```
gmarty: It's 204 as of yesterday. See https://github.com/fxbox/foxbox/pull/426/files.

tests/unit/lib/foxbox/services_test.js
gmarty: Yes, we need code coverage to know where to focus on writing new unit tests in the future.

tests/unit/test-utils.js
gmarty: I don't have any opinion either. Let's see how this works for us as we add more tests.

app/js/lib/foxbox/services.js
```diff
@@ -0,0 +1,273 @@
...
+
+  /**
+   * Retrieves the list of the services available. Relies on IndexedDB and
+   * in-memory cache.
+   *
+   * @return {Promise<Array<BaseService>>} A promise that resolves with an
+   * array of objects.
+   */
+  getAll() {
+    return this[p.getCache]().then((cache) => Array.from(cache.values()));
```
gmarty: To properly tackle #99, I'd like to have some kind of sorting here. Maybe sort by type (lights, camera, sensors...) then by name... anything that'd make listing order consistent and predictable.
What do you think?

app/js/lib/foxbox/foxbox.js
gmarty: The more I think about it, the less I like the `-changed` suffix. An event means a change happened, so that sounds like a redundancy to me. Also, we should take inspiration from the browser event, in this particular case, `online` seems appropriate.
Likewise, in `message-received`, a message related event is always going to be about receiving a message, so `message` looks good enough to me.

I know I suggested preterite verb, but maybe name only event names are better.
Do you have any preferences on that?

app/js/lib/foxbox/network.js
gmarty: I really like this, it makes it easy to read.

app/js/lib/foxbox/foxbox.js
azasypkin: >> I know I suggested preterite verb, but maybe name only event names are better.

:) Yeah, I have the same feeling honestly, just followed what we agreed on previously ;) So, remove ```preterite verb`` part and we're good? I'm fine with this.

>> Likewise, in message-received, a message related event is always going to be about receiving a message, so message looks good enough to me.

Not always :) In message app we have ```message-received``` ```message-sent``` ```message-failed-to-sent``` :) But in this case just ```message``` should be ok.

app/js/lib/foxbox/settings.js
gmarty: Settings essentially started as a wrapper around local storage. I still think that it's a good thing to have all local storage elements managed in the same class. But I understand your point of view.

app/js/views/tag-item.jsx
azasypkin: Sure, will have something like ```console.error('Adding new tags is not supported!')```

gmarty:
I did the review (except the tests), but because it's a big PR I want to make sure I take the time to fully understand the changes and what it means for the architecture of the lib.
I'll do that and check the tests this afternoon.

But so far so good. That's a great job!

azasypkin:
> I did the review (except the tests), but because it's a big PR I want to make sure I take the time to fully understand the changes and what it means for the architecture of the lib.
I'll do that and check the tests this afternoon.

Sure, take your time and thanks for the review! I won't touch this PR to until you're fully done so that you have original "snapshot".

app/js/lib/foxbox/foxbox.js
gmarty: I think we need an event to inform the view that a new box was detected or a previous box was disconnected.
Something like `boxes` (on the same pattern than `services` events). We can then make sure that the view is always in sync with whatever the registration server returns.

app/js/lib/foxbox/services.js
azasypkin: Well I really have a mixed feelings about ```{Object}``` here, but can't find anything JS specific to advocate myself - so will make it ```{Object}``` :)

tests/foxbox-simulator/http-server.js
azasypkin: Noted, will change!

app/js/lib/foxbox/foxbox.js
azasypkin: >> the view is always in sync with whatever the registration server returns.

You mean, login view, right? If so, yes, let me fix that.

app/js/lib/foxbox/foxbox.js
azasypkin: I think we can try to not be very strict right now, and just follow "lower case + dashes to separate words + common sense" rule and see how it goes :)

app/js/lib/foxbox/foxbox.js
azasypkin: So, the easiest way (to not make this PR even more bigger :)) is to introduce ```discovery``` event and call it every time registration server returns a valid response (if we override ```this[p.boxes]``` I'll fire event). Later on we can improve logic to check if box list has actually changed.

app/js/lib/foxbox/services.js
azasypkin: Ideally yes.In #99 my goal was to remove that annoying re-ordering that happens all the time :) So yeah, the thing that annoyed me is fixed with this PR, and we can use #99 to implement smarter sorting when we know for sure what sorting order we want.

app/js/lib/foxbox/settings.js
azasypkin: Filed https://github.com/fxbox/app/issues/130

tests/unit/lib/foxbox/services_test.js
```diff
@@ -0,0 +1,195 @@
...
+    dbStub.getServices.returns(Promise.resolve(dbServices));
+
+    apiStub = sinon.stub({ get: () => {} });
+
+    settingsStub = sinon.stub({ pollingInterval: 2000 });
+
+    services = new Services(dbStub, apiStub, settingsStub);
+  });
+
+  it('getAll returns all currently cached services', function(done) {
```
gmarty: nit: you're using double quotes around method names in the api test file above.

app/js/lib/foxbox/services.js
```diff
@@ -0,0 +1,273 @@
...
+
+            if (!fetchedService) {
+              this[p.db].deleteService(storedService.data);
+              cache.delete(storedService.data.id);
+            }
+          });
+        }
+
+        if (servicesToAddCount > 0 || servicesToRemoveCount > 0) {
+          this.emit('services-changed');
```
gmarty: It's not part of this patch, but I wonder if we should consolidate the way `services-changed` and `service-changed` events work. One send data and the other doesn't. Maybe they both should.

gmarty:
OK, I left a couple of nit/question but it looks good to me. r+

app/js/lib/foxbox/services.js
azasypkin: Yep, we should discuss this in the followups. But I'd rather have ```service-added``` and ```service-removed``` instead of ```services-changed```. Yes it can be triggered a plenty of times consequently, but home view should just throttle it like it would do with any other frequent event (scrolling and etc.).

PR #121, opened 2016-04-22
Post-install instructions are not needed. r=azasypkin

gmarty:


gmarty:
The post-install hack is not needed anymore (I had to clear the cache on Travis).
@azasypkin: It's probably safe to land, but can you take a look?

azasypkin:
I can't think of any reason why may need this.... so land it!

PR #123, opened 2016-04-25
Gulp: Make sure we tear down servers and simulators

### LINKED ISSUE #122: Selenium tests hang when a webpage can't be loaded
@gmarty spotted this odd behavior https://travis-ci.org/fxbox/app/builds/125043379#L426

In order to repro, apply this patch on top of ee6997db0676c079b10d709ba7540e99cdd64d7a and launch `npm test`. 

I suspect [that sequence](https://github.com/fxbox/app/blob/d25a865047e62f2c3c2de9d8e737b65077fcfeae/gulpfile.js#L242) to not run `stop-simulator` when the previous task fails. Hence, a subprocess would still lives, that why it hangs.

JohanLorenzo:
Fixes #122

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -241,9 +241,10 @@ gulp.task('run-test-integration', function() {
 });
 
 gulp.task('test-integration', function(cb) {
-  return runSequence(
-    'start-simulators', 'run-test-integration', 'stop-simulators', cb
-  );
+  runSequence('start-simulators', 'run-test-integration', () => {
+    // Tear down whatever the result is
+    runSequence('stop-simulators', cb);
+  });
 });
 
 gulp.task('run-test-e2e', function() {
@@ -257,8 +258,9 @@ gulp.task('test', function(cb) {
     'run-unit-tests',
     'webserver',
     'test-integration',
-    'stop-webserver',
-    cb
+    () => { // Tear down whatever the result is
+      runSequence('stop-webserver', cb);
+    }
   );
 });
 


JohanLorenzo:
I found a way to run some tasks even though the previous failed. In order to show it works, here are:

* [A patch](https://github.com/fxbox/app/pull/123/commits/c68cd976b5826cfaed3e01ae6ebc74c574aa1512) that makes the [integration tests hanging](https://travis-ci.org/fxbox/app/builds/125512278#L425).
* The fix on top the error => the error is [immediately reported](https://travis-ci.org/fxbox/app/builds/125512980#L413). 
* The fix alone, with the [tests passing](https://travis-ci.org/fxbox/app/builds/125513496). 

r? @gmarty

gulpfile.js
```diff
@@ -241,9 +241,9 @@ gulp.task('run-test-integration', function() {
 });
 
 gulp.task('test-integration', function(cb) {
-  return runSequence(
-    'start-simulators', 'run-test-integration', 'stop-simulators', cb
-  );
+  runSequence('start-simulators', 'run-test-integration', function tearDown() {
```
JohanLorenzo: I know, the function can remain anonymous, but I find the intent of wrapping the callback clearer with that kind of names

gmarty: nit: can you use arrow function whenever possible?

gulpfile.js
```diff
@@ -257,8 +257,9 @@ gulp.task('test', function(cb) {
     'run-unit-tests',
     'webserver',
     'test-integration',
-    'stop-webserver',
-    cb
+    function tearDown() {
+      runSequence('stop-webserver', cb);
+    }
```
gmarty: nit: same here.

gmarty:
I prefer arrow functions, specially in a Gulp file, over named functions. Maybe add a comment to explain why you're doing it.
But r+.

JohanLorenzo:
I avoided to use arrow function for a matter of consistency. None is used within the entire gulpfile. I don't mind using them, though. I updated my PR and added comment in the code.

PR #127, opened 2016-04-25
Fix the login page UI. r=azasypkin

gmarty:
Fix the multibox login page that was broken for some reason.

### DIFF
--- a/app/js/lib/foxbox/foxbox.js
+++ b/app/js/lib/foxbox/foxbox.js
@@ -53,15 +53,16 @@ const isSimilar = (objectA, objectB) => {
 };
 
 export default class Foxbox extends Service {
-  constructor() {
+  constructor({ settings, db, net } = {}) {
     super();
 
     // Private properties.
-    this[p.settings] = new Settings();
-    this[p.db] = new Db();
-    this[p.net] = new Network(this[p.settings], (foxboxOnline) => {
-      this._dispatchEvent('box-online', foxboxOnline);
-    });
+    this[p.settings] = settings || new Settings();
+    this[p.db] = db || new Db();
+    this[p.net] = net ||
+      new Network(this[p.settings], (foxboxOnline) => {
+        this._dispatchEvent('box-online', foxboxOnline);
+      });
     this[p.boxes] = Object.freeze([]);
     this[p.isPollingEnabled] = false;
     this[p.nextPollTimeout] = null;
@@ -127,8 +128,8 @@ export default class Foxbox extends Service {
     return this[p.settings].localOrigin;
   }
 
-  get clientId() {
-    return this[p.settings].clientId;
+  get client() {
+    return this[p.settings].client;
   }
 
   get boxes() {
@@ -164,15 +165,14 @@ export default class Foxbox extends Service {
           boxes
             .filter((box) => box.timestamp - now >= 0)
             .map((box) => {
-                // NOTE(sgiles): There is consideration to allow
-                // only "local_origin" and "tunnel_origin", removing the
-                // need to parse message - this merges the relevant message
-                // fields into the main object
-                const { local_origin, tunnel_origin } =
-                  JSON.parse(box.message);
-                box.local_origin = local_origin;
-                box.tunnel_origin = tunnel_origin;
-                return Object.freeze(box);
+              // NOTE(sgiles): There is consideration to allow
+              // only "local_origin" and "tunnel_origin", removing the
+              // need to parse message - this merges the relevant message
+              // fields into the main object
+              const { local_origin, tunnel_origin } = JSON.parse(box.message);
+              const client = box.client;
+
+              return Object.freeze({ local_origin, tunnel_origin, client });
             })
         );
 
@@ -238,7 +238,7 @@ export default class Foxbox extends Service {
       this[p.settings].tunnelOrigin = '';
     }
 
-    this[p.settings].clientId = box.client;
+    this[p.settings].client = box.client;
     this[p.settings].configured = true;
   }
 
--- a/app/js/views/user-login.jsx
+++ b/app/js/views/user-login.jsx
@@ -8,13 +8,15 @@ export default class UserLogin extends BaseView {
 
     this.state = {
       boxes: props.foxbox.boxes,
-      value: null,
+      selectedBox: null,
       loginEnabled: false,
     };
 
     this.foxbox = props.foxbox;
 
     this.onBoxOnline = this.onBoxOnline.bind(this);
+    this.onSelectChange = this.onSelectChange.bind(this);
+    this.onFormSubmit = this.onFormSubmit.bind(this);
   }
 
   componentDidMount() {
@@ -25,21 +27,21 @@ export default class UserLogin extends BaseView {
     this.foxbox.removeEventListener('box-online', this.onBoxOnline);
   }
 
-  handleOnChange(evt) {
-    const value = evt.target.selectedIndex;
+  onSelectChange(evt) {
+    const selectedBox = evt.target.selectedIndex;
 
-    this.setState({ value });
-    this.foxbox.selectBox(value);
+    this.setState({ selectedBox });
+    this.foxbox.selectBox(selectedBox);
   }
 
-  handleOnSubmit(evt) {
+  onFormSubmit(evt) {
     evt.preventDefault(); // Avoid redirection to /?.
 
     this.foxbox.login();
   }
 
-  onBoxOnline(online) {
-    this.setState({ loginEnabled: online });
+  onBoxOnline(loginEnabled) {
+    this.setState({ loginEnabled });
   }
 
   renderHeader() {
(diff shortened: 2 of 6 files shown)


app/js/lib/foxbox/foxbox.js
```diff
@@ -164,15 +164,14 @@ export default class Foxbox extends Service {
...
-                box.tunnel_origin = tunnel_origin;
-                return Object.freeze(box);
+              // NOTE(sgiles): There is consideration to allow
+              // only "local_origin" and "tunnel_origin", removing the
+              // need to parse message - this merges the relevant message
+              // fields into the main object
+              const { local_origin, tunnel_origin } = JSON.parse(box.message);
+              const client = box.client;
+
+              return Object.freeze({ local_origin, tunnel_origin, client });
```
gmarty: Let's only keep what we're actually using and ignore the rest.

azasypkin:
Sorry, was busy entire evening with unit tests for my PR, so didn't have time to look at your PR, will look at it as the first thing tomorrow!

In the meantime, it looks like a great chance to add few tests for the foxbox discovery mechanism and login page, would you mind to add some?

app/js/lib/foxbox/foxbox.js
azasypkin: Makes sense!

app/js/views/user-login.jsx
```diff
@@ -7,14 +7,15 @@ export default class UserLogin extends BaseView {
     super(props);
 
     this.state = {
-      boxes: props.foxbox.boxes,
       value: null,
```
azasypkin: Not related to your PR, but ```value``` name is so vague :)

app/js/views/user-login.jsx
```diff
@@ -51,33 +52,34 @@ export default class UserLogin extends BaseView {
...
   renderBody() {
-    let boxes = (<div hidden></div>);
+    let boxNodes = null;
 
-    if (this.state.boxes.length > 1) {
-      let value = this.state.value !== null ? 0 : this.state.value;
-      const options = this.state.boxes.map((box, index) => {
-        if (box.clientId === this.foxbox.clientId) {
+    if (this.foxbox.boxes.length > 1) {
+      let value = (!this.state.value) ? 0 : this.state.value;
```
azasypkin: __optional nit:__ round brackets around ```!this.state.value``` feel redundant.

app/js/lib/foxbox/foxbox.js
azasypkin: __note:__ not really an issue, but now we have ```box.client``` and ```foxbox.clientId``` that are essentially the same thing... But I'm fine if you feel it's OK.

azasypkin:
Looks good to me, r+, just few minor and mostly optional nits. But still it would be great if you can add few unit tests for the affected pieces.

app/js/views/user-login.jsx
azasypkin: __note:__ Follow up from here https://github.com/fxbox/app/pull/120/files#r61107814

If we want box list to be dynamically updated, we should keep ```boxes``` in the state, so that when box list changes we change state and then React re-renders view.

app/js/lib/foxbox/foxbox.js
gmarty: I'll rename all to `client`.

app/js/views/user-login.jsx
gmarty: I thought about that too, let's rename it.

app/js/views/user-login.jsx
gmarty: Ah yeah, I'll revert that.

tests/unit/lib/foxbox/foxbox_test.js
```diff
@@ -0,0 +1,84 @@
...
+      client: 'def',
+      message: JSON.stringify({
+        local_origin: 'http://local.def.box.knilxof.org:3000',
+        tunnel_origin: 'null',
+      }),
+      timestamp: Math.floor(Date.now() / 1000),
+    },
+  ]);
+
+  function getMockNet(single = true) {
```
gmarty: We'll need a better mock and in its own file, but I wanted to keep this patch small for now.

tests/unit/views/user-login_test.js
```diff
@@ -0,0 +1,64 @@
...
+    }),
+    timestamp: Math.floor(Date.now() / 1000),
+  },
+]);
+
+describe('User login view tests', function() {
+  let foxboxStub, component, renderedDOM;
+
+  function getFoxboxStub(single = true) {
+    foxboxStub = sinon.stub({
```
gmarty: We should also reuse the net mock here when it's an independent file.

app/js/lib/foxbox/foxbox.js
```diff
@@ -53,15 +53,15 @@ const isSimilar = (objectA, objectB) => {
 };
 
 export default class Foxbox extends Service {
-  constructor() {
+  constructor({ settings, db, net } = { settings: null, db: null, net: null }) {
```
azasypkin: __question:__ do we really need this default null values?

app/js/lib/foxbox/foxbox.js
```diff
@@ -53,15 +53,15 @@ const isSimilar = (objectA, objectB) => {
...
 
     // Private properties.
-    this[p.settings] = new Settings();
-    this[p.db] = new Db();
-    this[p.net] = new Network(this[p.settings], (foxboxOnline) => {
-      this._dispatchEvent('box-online', foxboxOnline);
-    });
+    this[p.settings] = settings || new Settings();
+    this[p.db] = db || new Db();
+    this[p.net] = net || new Network(this[p.settings], (foxboxOnline) => {
```
azasypkin: __nit:__ weird indent I'd rather do either
```
this[p.net] = net || new Network(this[p.settings], (foxboxOnline) => {
  this._dispatchEvent('box-online', foxboxOnline);
});
```
or
```
this[p.net] = net || 
  new Network(this[p.settings], (foxboxOnline) => {
    this._dispatchEvent('box-online', foxboxOnline);
  });
```

And yeah we should make our IDE's to format in a similar way :)

tests/unit/lib/foxbox/foxbox_test.js
azasypkin: >  in its own file,

Honestly I don't really like mocks in a separate files like some used to do in gaia. Imo using sinon is a way more flexible and easy....

Moreover how do you feel about something more sinon-y and unit-testy :) Like this

```
 let netStub;

  beforeEach(function() {
    netStub = sinon.stub({
      init: () => {},
      fetchJSON: () => {}
    });
  });
  
  it/before/any-block-you-need(..., function() {
    netStub.fetchJSON.withArgs(...guard yourself with correct args.).returns(multiBox);
    
    // or 

    netStub.fetchJSON.withArgs(...guard yourself with correct args.).returns(singleBox);
  });
```

tests/unit/lib/foxbox/foxbox_test.js
```diff
@@ -0,0 +1,84 @@
...
+
+      foxbox.init()
+        .then(() => {
+          sinon.assert.calledOnce(mockNet.init);
+        })
+        .then(done, done);
+    });
+  });
+
+  describe('_initDiscovery >', function() {
```
azasypkin: Ideally (always?) we don't want to test private methods as they are not part of the public API and can be refactored in any way and tests should not change if input/output is the same. It feels like what we really want to test here is public ```foxbox.init``` and specifically that:
* it calls ```fetchJSON``` with correct arguments;
* if it fails and we don't have local/remote origin from the previous time, it's called once again after interval (can be mocked with sinon.clock);
* if it succeed - box list is updated (unfortunately I've added this event in my PR that is waiting your one to land :), with this event it would be easier, but you can rely on box-online event for now);
* user session is initialized (it's even easier, but maybe not in this PR).

I know it's quite painful to test promise based code (especially when methods don't return promises that we'd like to test). We have two helper methods that should help with this - https://github.com/azasypkin/app/blob/issue-111-watchers/tests/unit/test-utils.js

What do you think?

tests/unit/views/user-login_test.js
```diff
@@ -0,0 +1,64 @@
...
+
+    component = TestUtils.renderIntoDocument(<UserLogin foxbox={foxboxStub}/>);
+    renderedDOM = () => ReactDOM.findDOMNode(component);
+
+    const selects = renderedDOM().querySelectorAll('select');
+    assert.equal(selects.length, 0);
+  });
+
+  it('Has a box selector when more than one box is found', function() {
+    getFoxboxStub(false);
```
azasypkin: __nit:__ it's not really obvious what this ```false``` means. Why not just modify boxes list in the test (or in before if it was ```describe``` block)? Or at least mark it some how ```getFoxboxStub(false /* singleBox */);```

tests/unit/views/user-login_test.js
```diff
@@ -0,0 +1,64 @@
...
+
+    component = TestUtils.renderIntoDocument(<UserLogin foxbox={foxboxStub}/>);
+    renderedDOM = () => ReactDOM.findDOMNode(component);
+
+    const selects = renderedDOM().querySelectorAll('select');
+    assert.equal(selects.length, 1);
```
azasypkin: __nit:__ can we also test ```option``` elements, to make sure box addresses are displayed correctly

__nit:__ you can use ```assert.lengthOf```

tests/unit/views/user-login_test.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  it('Has no box selector when only one box is found', function() {
+    getFoxboxStub();
+
+    component = TestUtils.renderIntoDocument(<UserLogin foxbox={foxboxStub}/>);
+    renderedDOM = () => ReactDOM.findDOMNode(component);
+
+    const selects = renderedDOM().querySelectorAll('select');
+    assert.equal(selects.length, 0);
```
azasypkin: __nit:__ you can use ```assert.lengthOf```

app/js/lib/foxbox/foxbox.js
gmarty: Yes, we don't need the null values.

app/js/lib/foxbox/foxbox.js
gmarty: OK, I'll fix it.

tests/unit/lib/foxbox/foxbox_test.js
gmarty: Yep, I agree, we actually want to test `init` here.
Regarding the test plan I also agree but that's out of the scope of this PR. I simply want to make sure that we don't regress with the code related to the registration server and the box discovery for now.

PR #128, opened 2016-04-25
When using the indexedDB, stop wrapping services info in data object. r=azasypkin

gmarty:
Wrapping info in object (e.g. `{ id, data }`) when interacting with the indexedDB only caused confusion. That makes the signature of `foxbox#getServices()` and `foxbox#[p.fetchServices]()` the same.

@azasypkin: What do you think?

### DIFF
--- a/app/js/lib/foxbox/db.js
+++ b/app/js/lib/foxbox/db.js
@@ -65,7 +65,6 @@ export default class Db {
     if (fromVersion < 1) {
       let store = db.createObjectStore(DB_SERVICE_STORE, { keyPath: 'id' });
       store.createIndex('id', 'id', { unique: true });
-      store.createIndex('type', 'type', { unique: false });
 
       store = db.createObjectStore(DB_TAG_STORE, {
         keyPath: 'id',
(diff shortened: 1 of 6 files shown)


app/js/lib/foxbox/db.js
```diff
@@ -65,7 +65,6 @@ export default class Db {
     if (fromVersion < 1) {
       let store = db.createObjectStore(DB_SERVICE_STORE, { keyPath: 'id' });
       store.createIndex('id', 'id', { unique: true });
-      store.createIndex('type', 'type', { unique: false });
```
gmarty: I didn't write an upgrade script. This means users will have to call `foxbox.clear()` when this patch lands.
Let me know if you think upgrade should be considered.

azasypkin: > I didn't write an upgrade script. This means users will have to call foxbox.clear() when this patch lands.
Let me know if you think upgrade should be considered.

Nah, I think we don't need to bother about upgrade at this stage. At such early stage I always try to not introduce sophisticated code that isn't critically necessary - it will just increase maintenance cost for us.

> This also means that the data must have either a unique id property or no id property at all (in which case, it'll be created depending on the store configuration).

Sounds reasonable.

azasypkin:
The second commit (the one with IndexedDB changes) looks good to me, r+. If it doesn't depend on https://github.com/fxbox/app/pull/127, I think we can safely land it now. It will introduce some conflicts for my PR you're reviewing - but those are small, I'll resolve it and test once you land your changes.

Issue #111, opened 2016-04-21
[Experiment] Implement Watchers


azasypkin:
We don't have websockets/event source yet, but we need to display changes in state (door lock closed manually, user switched lights off and etc) - UI should be updated accordingly. The only way for now is to do polling - with the Taxonomy we can do very efficient polling.

So in the scope of this bug I want to try idea with Watchers that could accept any number of getter and poll all of them at once.

Likely this experiment will rely on https://github.com/fxbox/app/issues/107

azasypkin:
Depends on https://github.com/fxbox/app/pull/115

azasypkin:
Depends on https://github.com/fxbox/app/issues/116

azasypkin:
Fixed via https://github.com/fxbox/app/pull/129

PR #129, opened 2016-04-26
Implement Watchers (fixes #107, #111). r=gmarty

### LINKED ISSUE #107: Start using EventDispatcher for Foxbox and Services

azasypkin:


azasypkin:
Hey @gmarty r?, I think the code part is ready for review, could you please take a look?

I'm just planning to add few more tests.

app/js/lib/foxbox/network.js
```diff
@@ -183,19 +186,26 @@ export default class Network extends EventDispatcher {
...
     if (this[p.settings].tunnelOrigin) {
-      this[p.ping](`${this.tunnelOrigin}/ping`)
-        .then((isOnline) => this[p.onPong](previousState, p.remote, isOnline));
+      pingPromises.push(
+        this[p.ping](`${this.tunnelOrigin}/ping`)
+          .then((isOnline) => this[p.onPong](previousState, p.remote, isOnline))
+      );
     }
+
+    return Promise.all(pingPromises);
```
gmarty: I think we have a similar issue here than when trying to put these 2 pings in a promise.
If one of the ping takes very long to resolve, the next ping will be delayed for as long. Maybe it's better to have 2 timers here: 1 for local and 1 for remote.
What do you think?

app/js/lib/foxbox/sequential-timer.js
```diff
@@ -0,0 +1,94 @@
+const p = Object.freeze({
```
gmarty: `use strict` please :-)

app/js/lib/foxbox/sequential-timer.js
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Creates new SequentialTimer instance.
+   * @param {number} interval Minimum interval between two consequent ticks.
+   */
+  constructor(interval) {
+    this.interval = interval;
+
+    this[p.isStarted] = false;
+    this[p.nextTickHandle] = null;
+    this[p.onTick] = null;
```
gmarty: Can we seal or freeze the instance?

app/js/lib/foxbox/sequential-timer.js
```diff
@@ -0,0 +1,94 @@
...
+          console.error(
+            'onTick handler failed, scheduling one more attempt anyway: %o',
+            error
+          );
+        })
+        .then(() => {
+          this[p.nextTickHandle] = null;
+          this[p.scheduleTick]();
+        });
+    }, this.interval);
```
gmarty: That's super neat. I love this promise based timer.

app/js/lib/foxbox/network.js
azasypkin: > I think we have a similar issue here than when trying to put these 2 pings in a promise.
If one of the ping takes very long to resolve, the next ping will be delayed for as long. 

It's hard to say, preliminary ```ping takes very long to resolve``` should be an edge case, and more important to have less network requests as possible.

> Maybe it's better to have 2 timers here: 1 for local and 1 for remote.
What do you think?

Yeah, maybe we can try that, I'm wondering maybe they should even have different periods... Do you want me to try to do it in this PR or file appropriate issue?

app/js/lib/foxbox/sequential-timer.js
azasypkin: Yeah, I'll ```seal``` it

app/js/lib/foxbox/sequential-timer.js
azasypkin: Ha-ha, will add!

app/js/lib/foxbox/sequential-timer.js
azasypkin: Thanks! I'm glad that you like it :)

app/css/components/service-list.css
```diff
@@ -4,11 +4,7 @@
 
 .service-list__item {
   display: flex;
-  padding: 0;
   position: relative;
-}
-
-.service-list__item[data-icon] {
```
azasypkin: It seems we don't and won't have service list items without icons or I am missing something?

app/css/views/services-list.css
```diff
@@ -6,10 +6,28 @@
...
+  5%, 95% {
+    background-size: 2.3rem;
+  }
+
+  10%, 85%, 100% {
+    background-size: 2.5rem;
+  }
+}
+
+.motion-sensor-item {
```
azasypkin: Trying to use more BEMy approach.

app/css/views/services-list.css
```diff
@@ -6,10 +6,28 @@
...
   background-image: url('services-list/motion-sensor.svg');
 }
 
-[data-icon="door-lock"] {
-  background-image: url('services-list/door-lock.svg');
+.motion-sensor-item--motion-detected {
```
azasypkin: It's just something for intermediate period to see that sensor is activated. But at some point we'll definitely need to implement all these shiny icons provided by VD.

app/js/lib/foxbox/api.js
```diff
@@ -148,4 +222,78 @@ export default class API {
...
+    // so let's return early in this case.
+    if (this[p.watchGetters].size === 0) {
+      return Promise.resolve();
+    }
+
+    const selectors = Array.from(this[p.watchGetters].values()).map(
+      ({ id }) => ({ id })
+    );
+
+    return this.put('channels/get', selectors)
```
azasypkin: That's a great feature of taxonomy we can group all watched getters into one request!

app/js/lib/foxbox/api.js
```diff
@@ -148,4 +222,78 @@ export default class API {
...
+   *
+   * @param {{ id: string, value: Object }} getter Getter to update value for.
+   * @param {Object} getterValue Getter value returned from the server.
+   *
+   * @private
+   */
+  [p.updateGetterValue](getter, getterValue) {
+    let valueChanged = false;
+
+    if (!getterValue || !getter.value) {
```
azasypkin: That's a special case when state of the value is not known yet (happens for OpenZWave pretty frequently)

app/js/lib/foxbox/api.js
```diff
@@ -148,4 +222,78 @@ export default class API {
...
+        // If value is a non-null object, we use their JSON representation to
+        // compare values.
+        valueChanged = JSON.stringify(newValue) !== JSON.stringify(oldValue);
+      } else {
+        valueChanged = newValue !== oldValue;
+      }
+    }
+
+    if (valueChanged) {
+      getter.value = Object.freeze(getterValue);
```
azasypkin: I keep entire value container (e.g. ```{ OpenClosed: "Open" }``` instead of just ```"Open"```) for two reasons:
* To have the same event object format as returned value in ```BaseService.get```;
* To be more restrictive in comparing old and new value to make sure type of the value hasn't changed (it's not 100% protection, but still something).

app/js/lib/foxbox/services.js
```diff
@@ -138,91 +136,66 @@ export default class Services extends EventDispatcher {
...
-              cache.delete(storedService.id);
-            }
-          });
-        }
-
-        if (servicesToAddCount > 0 || servicesToRemoveCount > 0) {
-          this.emit('services-changed');
-        }
-      });
+      if (servicesToAddCount > 0 || servicesToRemoveCount > 0) {
```
azasypkin: As I said before, eventually I'd like us to generate two events instead of this one - ```service-added``` with new service as argument and ```service-removed``` with removed service as argument - so that view can unsubscribe from its event handlers safely even still being active and let user know that currently opened service has disappeared.

app/js/lib/foxbox/settings.js
```diff
@@ -187,6 +188,14 @@ export default class Settings extends EventDispatcher {
     return ONLINE_CHECKING_LONG_INTERVAL;
   }
 
+  /**
+   * Minimal interval between consequent value watcher requests.
+   * @return {number}
+   */
+  get watchInterval() {
```
azasypkin: If this watcher idea sticks, we can later implement new setting to enable/disable watching, but good thing that in most common conditions like ```box is not available``` or ```user is not authenticated``` our current API won't try to make any call anyway.

app/js/views/services-list-item.jsx
```diff
@@ -8,10 +8,13 @@ export default class ServicesListItem extends React.Component {
       available: false,
       on: true,
       locked: true,
+      motionDetected: false,
```
azasypkin: Again if this idea sticks, next step for us would be to finally split this view into several service specific mini-views. It's going to grow and become a total mess :)

tests/unit/lib/foxbox/sequential-timer_test.js
```diff
@@ -0,0 +1,126 @@
+import { waitForNextMacroTask } from '../../test-utils';
```
azasypkin: Repeat myself - testing promise based code when promises are not exposed via public api is really __hard__ :/ If browser supports latest Micro and Macro task specs (like browsers we support) we can rely on ```waitForNextMacroTask```

tests/unit/lib/foxbox/sequential-timer_test.js
```diff
@@ -0,0 +1,126 @@
...
+        .then(() => {
+          // Should not be called earlier than needed.
+          sinon.assert.notCalled(onTick);
+          this.sinon.clock.tick(1);
+
+          return waitForNextMacroTask();
+        })
+        .then(() => {
+          sinon.assert.calledOnce(onTick);
+          this.sinon.clock.tick(DEFAULT_INTERVAL + 10);
```
azasypkin: __note to myself:__ here should be just ```DEFAULT_INTERVAL```

app/js/lib/foxbox/sequential-timer.js
gmarty: We start to have a lot of different files in `app/js/foxbox` and I think we need some sort of organisation.
Maybe we can start by placing all the base classes that other classes are extending in a subfolder (`base` or `extends`...). What do you think?

app/css/components/service-list.css
gmarty: We have a default icon for unknown services which is a question mark. So yes, all services have an icon.

app/css/views/services-list.css
gmarty: Yes, I prefer not to spend too much time on the design/UX while the scope of the project is still not defined.

app/js/lib/foxbox/api.js
```diff
@@ -148,4 +222,78 @@ export default class API {
...
+        return;
+      }
+
+      const newValue = getterValue[valueKind];
+      const oldValue = getter.value[valueKind];
+
+      if (newValue && oldValue && typeof newValue === 'object') {
+        // If value is a non-null object, we use their JSON representation to
+        // compare values.
+        valueChanged = JSON.stringify(newValue) !== JSON.stringify(oldValue);
```
gmarty: I'm sure there are more efficient method to do deep object comparison. Also if the properties order is not the same, the comparison will fail. As per the es spec, the objects property order shouldn't be relied on.

app/js/lib/foxbox/sequential-timer.js
azasypkin: Yes, what about ```shared``` like react does [1] or ```common``` like in angular [2]?

Honestly I much more like how angular is organized in general comparing to react, so ```common``` sounds better to me, but I'll choose whatever you prefer.

[1] https://github.com/facebook/react/tree/master/src/shared
[2] https://github.com/angular/angular/tree/master/modules/angular2/src/common

app/js/lib/foxbox/api.js
azasypkin: Definitely, the main reason why I don't really care much - we don't have such values currently as far as I know, so I didn't want to spend time on something that we don't use - just a basic placeholder. If one day we have such values (or when we have more time) - I think we can reuse ```isSimilar``` method from ```services (it will be moved to a common place and maybe improved even more).

What do you think?

app/js/lib/foxbox/network.js
gmarty: Yes please can you try to do that in that PR?
Also different periods of time sound good.

Maybe also we should use the connection type from the net-info API (only in Chrome for now). If on Wifi we can ping more often that on radio. This optimisation can definitely wait for another PR as long as we fill a bug for it.

app/js/lib/foxbox/sequential-timer.js
```diff
@@ -0,0 +1,94 @@
...
+    this[p.onTick] = null;
+  }
+
+  /**
+   * Indicates whether timer started or not.
+   *
+   * @return {boolean}
+   */
+  get isStarted() {
+    return this[p.isStarted];
```
gmarty: I would have called it `hasStarted`, but maybe `started` is good enough: quick and easy to remember.

azasypkin: Okay, let's go with ```started```

app/js/lib/foxbox/services.js
```diff
@@ -67,8 +66,11 @@ export default class Services extends EventDispatcher {
     this[p.settings] = settings;
 
     this[p.cache] = null;
-    this[p.isPollingEnabled] = false;
-    this[p.nextPollTimeout] = null;
+    this[p.pollingTimer] = new SequentialTimer(
+      this[p.settings].pollingInterval
```
gmarty: In this case, you used `settings` directly instead of `this[p.settings]` in one of the files above.
Decide which one prefer and use it consistently.

app/js/lib/foxbox/services.js
gmarty: Yes that's a good idea. Maybe file an issue for that and let's do it later.

app/js/lib/foxbox/services.js
azasypkin: Haven't decided what is better yet :) But yeah agree, let's be consistent

app/js/lib/foxbox/services/motion-sensor.js
```diff
@@ -4,13 +4,68 @@ import BaseService from './base';
...
+
+/**
+ * Converts motion state value to boolean. Considers unknown state (null) the
+ * same as state when motion is not detected.
+ *
+ * @param {Object} motionState Motion state object.
+ * @return {boolean}
+ * @private
+ */
+function motionStateToBoolean(motionState) {
```
gmarty: In `app/js/lib/foxbox/services.js`, a similar function is declared using `const`.
Let's use a consistent pattern in both cases.

app/js/views/services-list-item.jsx
gmarty: More than agreed here. We need something similar to `service.jsx` or even share some logic between both views.

app/js/lib/foxbox/network.js
azasypkin: > Yes please can you try to do that in that PR?

Okay, trying.;

> Also different periods of time sound good.
Maybe also we should use the connection type from the net-info API (only in Chrome for now). If on Wifi we can ping more often that on radio. This optimisation can definitely wait for another PR as long as we fill a bug for it.

Done https://github.com/fxbox/app/issues/135

app/js/views/services-list-item.jsx
```diff
@@ -41,6 +49,14 @@ export default class ServicesListItem extends React.Component {
     }
   }
 
+  componentWillUnmount() {
+    switch (this.service.type) {
+      case 'motion-sensor':
+        this.service.off('motion', this.onMotion);
```
gmarty: Can we do something like `this.service.offAll('motion')` here?
Or even add a new method to `EventDispatcher` to remove all listeners of all events? That'd make this part shorter and make sure we don't forget to unlisten to anything.

gmarty: I checked and `offAll` already does this if no parameters are supplied. So what about just `this.service.offAll()` here?

azasypkin: > Can we do something like this.service.offAll('motion') here?

Mmm, imo it's a bad practice, we should not remove event listeners we didn't set on the object we don't own. Especially when service is placed in a global pool and who knows what other subview/service/controller/etc set its own listeners.

> make sure we don't forget to unlisten to anything.

Well, if this is case - then we have bug in our code :) And we should investigate and fix it rather than masking :)

What do you think?

app/js/lib/foxbox/api.js
gmarty: Yes I was thinking of reusing `isSimilar` too.
Sounds good to me for now. If only temporary, maybe add a comment to remember to take another look when values like that are used.

app/js/lib/foxbox/sequential-timer.js
gmarty: I don't really have an opinion, pick the one you like best.
For the moment only `eventDispatcher` will go in there, right?

azasypkin: Mmm I was thinking about ```event-dispatcher.js, defer.js, sequential-timer.js``` - they are all really common for all.

app/js/views/services-list-item.jsx
gmarty: OK I understand, leave it like this.

gmarty:
I'm very excited about this PR, it shows the real potential of a powerful and user-friendly API! I left some comments, but r+!

app/js/lib/foxbox/api.js
azasypkin: Ok, will add comment!

app/js/lib/foxbox/services.js
azasypkin: Yep filed https://github.com/fxbox/app/issues/137

app/js/lib/foxbox/services/motion-sensor.js
azasypkin: Okay, let's try with const.

gmarty: In this case `EventDispatcher` needs updating too.

azasypkin: Ah, trap! :) Will update EventDispatcher too.

tests/unit/views/services/camera_test.js
```diff
@@ -20,6 +19,11 @@ describe('Camera service view tests', function() {
     component = TestUtils.renderIntoDocument(
       <CameraServiceView service={serviceMock} foxbox={foxboxStub}/>
     );
+
+    assert.isDefined(
+      component,
+      'ReactDOM.render() did not render component (see issue #133).'
```
gmarty: Can you use the link, an issue number only can be ambiguous (is it a foxbox error? a React error?):
`https://github.com/fxbox/app/issues/133`

azasypkin:
Thanks a lot for review! Review comments handled and travis is green, merging!

PR #139, opened 2016-04-28
Fix notification action path. r=jlorenzo

azasypkin:


azasypkin:
@JohanLorenzo r?

JohanLorenzo:
That works for me! Thanks!

Issue #130, opened 2016-04-26
[Refactoring] Migrate foxbox/settings private fields to new symbol-based naming system


azasypkin:
+ required fields should be exposed via getters and setters (where appropriate).

PR #141, opened 2016-05-03
Smart Polling: do not poll when document is not visible, do not poll watch values when no listeners, poll services and watchers only on Home view, do not ping if other request were successful. r=gmarty

azasypkin:


app/js/lib/foxbox/api.js
```diff
@@ -224,6 +230,28 @@ export default class API {
   }
 
   /**
+   * Returns promise that is resolved once document becomes visible.
+   *
+   * @returns {Promise}
+   * @private
+   */
+  [p.onceDocumentVisible]() {
```
azasypkin: It seems that there is no need in doing any requests when document is not active.

app/js/lib/foxbox/box-link.js
```diff
@@ -0,0 +1,131 @@
+import EventDispatcher from './common/event-dispatcher';
```
azasypkin: So then name ```link``` comes from the fact that both ```local``` and ```tunnel``` origins lead to the same ```box```, so we have one ```box``` we're connected to and possibly two ```routes``` or ```links``` to it. Initially I had ```box-route```, but I liked ```box-link``` more - tell me if sounds confusing to you.

app/js/lib/foxbox/foxbox.js
```diff
@@ -57,14 +57,7 @@ export default class Foxbox extends Service {
     // polling and triggering box-online events with a boolean
     // indicating if we have access to box or not.
     this._initDiscovery()
-      .then(() => this[p.net].init())
```
azasypkin: I've moved service list polling management to view itself, so that it can manage it as it wants. Still there is one edge case - when we're in the Service view - currently we don't poll anything here and I was planning to leave it as is for this PR since we don't have any rich Service view where we need this feature.

app/js/lib/foxbox/foxbox.js
```diff
@@ -223,24 +216,22 @@ export default class Foxbox extends Service {
...
-      // Remove the session param from the current location.
-      searchParams.delete('session_token');
-      location.search = searchParams.toString();
+    // There is a session token in the URL, let's remember it.
+    // @todo Find a better way to handle URL escape.
+    this[p.settings].session = url.searchParams.get('session_token')
+      .replace(/ /g, '+');
 
-      // Throwing here to abort the promise chain.
-      throw(new Error('Redirecting to a URL without session'));
```
azasypkin: I removed throwing of error since I was so annoyed by constantly seeing it in the console log :) It seems that it's not really an error, but rather normal program flow. Tell me if you prefer to revert this change!

app/js/lib/foxbox/network.js
```diff
@@ -49,43 +45,38 @@ export default class Network extends EventDispatcher {
...
-      pingInterval = this[p.settings].onlineCheckingInterval;
+      this[p.linkPingInterval] = this[p.settings].onlineCheckingInterval;
     }
 
-    // @todo Settings should emit event when it changes "tunnelOrigin" or
-    // "localOrigin" setting and we should listen for this change and start or
-    // stop timers according to these settings.
-    this[p.localPingTimer] = new SequentialTimer(pingInterval);
-    this[p.localPingTimer].start(this[p.pingLocalBox]);
+    this[p.settings].on('local-origin', this[p.onLinkOriginChange]);
```
azasypkin: Once we implement re-discovery, we'll need this peace of code anyway.

app/js/lib/foxbox/network.js
```diff
@@ -49,43 +45,38 @@ export default class Network extends EventDispatcher {
...
-      return this.tunnelOrigin;
+    if (this[p.localLink]) {
+      return this[p.localLink].origin;
+    } else if (this[p.tunnelLink]) {
+      return this[p.tunnelLink].origin;
     }
 
-    console.error('The box is out of reach.');
-    return this.localOrigin;
+    throw new Error('Origin is not accessible');
```
azasypkin: I don't have strong opinion here, but it seems if anything tries to access origin before we have it we should throw

app/js/lib/foxbox/network.js
```diff
@@ -187,94 +180,76 @@ export default class Network extends EventDispatcher {
...
    *
-   * @return {Promise}
    * @private
    */
-  [p.pingLocalBox]() {
-    // @todo Find a better way to detect if a local connection is active.
-    if (!this[p.settings].localOrigin) {
-      return Promise.resolve();
+  [p.pingLinks]() {
+    if (this[p.localLink]) {
```
azasypkin: Forcibly ping boxes without relying on auto ping interval.

app/js/lib/foxbox/network.js
```diff
@@ -187,94 +180,76 @@ export default class Network extends EventDispatcher {
...
+   * @param {string?} origin Origin string, if "null" - origin is not available
+   * anymore.
    * @private
    */
-  [p.pingRemoteBox]() {
-    // @todo Find a better way to detect if a tunnel connection is active.
-    if (!this[p.settings].tunnelOrigin) {
-      return Promise.resolve();
+  [p.updateLink](symbol, origin) {
+    const link = this[symbol];
```
azasypkin: P.S. Not really the code I'm proud of, but don't have anything better in my mind for the moment :/

app/js/lib/foxbox/network.js
```diff
@@ -187,94 +180,76 @@ export default class Network extends EventDispatcher {
...
-    // 'online' state has changed.
-    if (this[localOrRemote] === isOnline) {
-      return;
-    }
-
-    const previousOnlineState = this[p.local] || this[p.remote];
-
-    this[localOrRemote] = isOnline;
+  [p.onLinkOnlineChange]() {
+    const onlineStatus = !!this[p.localLink] && this[p.localLink].online ||
```
azasypkin: The idea is to remove box link instance once we don't have origin for it anymore.

app/js/lib/foxbox/services/base.js
```diff
@@ -112,34 +114,72 @@ export default class BaseService extends EventDispatcher {
...
    * Setups value watcher for the getter matching specified selector.
    *
-   * @param {Object} selector Selector to match getter which value we would like
-   * to watch.
+   * @param {string} alias Watcher alias to match getter which value we would
+   * like to watch.
    * @param {function} handler Function to be called once getter value changes.
    */
-  watch(selector, handler) {
+  watch(alias, handler) {
```
azasypkin: That my attempt to have the following behavior:
* Automatically start watching only when consumer wants to watch (e.g. when view does ```motionSensorService.watch('motion', ....)```, so service constructor will no longer forcibly start watching.
* Automatically stop watching when consumer doesn't want to watch anymore (e.g. when view does ```motionSensorService.unwatch('motion', .....)```;
* When service is being destroyed we can remove all event listeners in one place and log warning about forgotten event listener.

To do this my idea is to:
* Define which watchers service provides, what alias can be used by consumer (view) and how raw getter value is processed before it's returned to the consumer. Like this:
```
super(props, api, undefined, new Map([
  ['motion', ['OpenClosed', p.onMotionStateChanged]],
]));
```

* So consumer just does ```motionSensorService.watch('motion', ....)``` and receives boolean instead of ```{ OpenClosed: "Open"}```

Definitely ```watch\unwatch``` is much less flexible comparing to ```on\off\once\offAll```, but I chose a tradeoff solution (if that watcher subsystem proves to be good, we can switch back to ```on/offf``` and friends, it just requires more effort). Let me know your thoughts!

app/js/lib/foxbox/settings.js
```diff
@@ -203,4 +187,33 @@ export default class Settings extends EventDispatcher {
...
+    this[p[key]] = value;
+
+    if (value !== null && value !== undefined) {
+      storage.setItem(`${PREFIX}${key}`, value);
+    } else {
+      storage.removeItem(`${PREFIX}${key}`);
+    }
+
+    this.emit('setting', { key, value });
+    this.emit(
```
azasypkin: My idea was to only have separate event for every setting, so I'm thinking about removing of ```this.emit('setting', { key, value });```.

app/js/lib/foxbox/settings.js
```diff
@@ -203,4 +187,33 @@ export default class Settings extends EventDispatcher {
...
+      key.replace(UPPER_CASE_REGEX, (part) => `-${part.toLowerCase()}`),
```
azasypkin: ```camelCase``` property name to ```dashed-lower-case``` event name.

azasypkin:
Hey @gmarty it's still WIP, but will be great to have your early feedback! So f?.

I've deployed this version to https://azasypkin.github.io/app/ if you want  to see PR in action.

app/js/lib/foxbox/foxbox.js
gmarty: I threw here to stop the execution of the script. There's no need to go further because we're redirecting.
But maybe we still need to continue the execution in case the box can't be reached anymore.
So I'm fine with this change.

app/js/lib/foxbox/network.js
gmarty: Sounds good to me.

app/js/lib/foxbox/services/base.js
gmarty: I like the watcher API and I like the we explicitly have to declare them.
As you say, let's experiment with them and see how it goes.

app/js/lib/foxbox/settings.js
```diff
@@ -203,4 +187,33 @@ export default class Settings extends EventDispatcher {
   get apiVersion() {
     return API_VERSION;
   }
+
+  clear() {
+    return new Promise((resolve) => {
+      this.session = null;
+      // @todo Remove only the items set here.
```
gmarty: Can we fix this todo? Now that we have the list of properties in an array, we can iterate through them and selectively delete them while preserving the others.

app/js/lib/foxbox/settings.js
```diff
@@ -5,8 +5,8 @@ import EventDispatcher from './common/event-dispatcher';
 // Prefix all entries to avoid collisions.
```
gmarty: This PR is big enough already so if you want some incremental landing, maybe you can separate the effort for #130 in a first PR. If not, I'm fine with reviewing it as part of this patch.

app/js/lib/foxbox/settings.js
gmarty: I think that's a good idea.
Also what about listening to `storage` event and emitting a corresponding event here. This way we would cover the case where the local storage is modified by a third party script (or the devtools) and the lib would keep running consistently. Is it not overkilling? What do you think?

app/js/lib/foxbox/box-link.js
gmarty: I prefer `route` but `link` is fine too.

gmarty:
I left a few comments, but globally I'm happy with the direction of this PR!

app/js/lib/foxbox/services/base.js
azasypkin: Nice, thanks! At this stage we have that luxury to experiment without breaking something deployed in production :)

app/js/lib/foxbox/settings.js
azasypkin: Right, I will move settings part into separate PR.

app/js/lib/foxbox/settings.js
azasypkin: mmm yeah we can, the only thing is that we'll start using these strings in several places, maybe keep them in a single place and use in all related pieces so that we don't misspell and not forget about all places to update when new setting is added or existing is renamed? Like this:
```
const settings = Object.freeze({
  CONFIGURED: { key: 'configured', type: 'boolean' },
  SKIP_DISCOVERY: { key: 'skipDiscovery', type: 'boolean' },
  SERVICE_POLLING_ENABLED: {
    key: 'servicePollingEnabled',
    type: 'boolean',
    defaultValue: DEFAULT_SERVICE_POLLING_ENABLED,
  },
  LOCAL_ORIGIN: { key: 'localOrigin' },
  TUNNEL_ORIGIN: { key: 'tunnelOrigin' },
  CLIENT: { key: 'client' },
  SESSION: { key: 'session' },
  PUSH_ENDPOINT: { key: 'pushEndpoint' },
  PUSH_PUB_KEY: { key: 'pushPubKey' },
});

export default class Settings extends EventDispatcher {
  ......
  init() {
    Object.keys(settings).forEach((settingName) => {
      const setting = settings[settingName];

      let defaultValue = setting.defaultValue;
      if (defaultValue === undefined) {
        defaultValue = setting.type === 'boolean' ? false : null;
      }

      const storedValue = storage.getItem(`${PREFIX}${setting.key}`);

      let value;
      if (storedValue === null) {
        value = defaultValue;
      } else {
        value = setting.type === 'boolean' ?
          storedValue === 'true' : storedValue;
      }

      this[p[setting.key]] = value;
    });
  }

  get [settings.CONFIGURED.key]() {
    return this[p[settings.CONFIGURED.key]];
  }

  set [settings.CONFIGURED.key](value) {
    this[p.updateSetting](settings.CONFIGURED.key, value);
  }

  clear() {
    return new Promise((resolve) => {
      Object.keys(settings).forEach((settingName) => {
        storage.removeItem(`${PREFIX}${settings[settingName].key}`);
      });
      resolve();
    });
  }
```

Or it looks too cryptic already?

app/js/lib/foxbox/settings.js
azasypkin: Yeah, why not, but it's a good candidate for a separate PR/issue as it will involve a bit of additional code (to parse name of the key to remove `PREFIX` part + to deal correctly with boolean strings). Could you please issue and I'll look into it once my wips are landed?

app/js/lib/foxbox/settings.js
gmarty: Yeah, it's more complex than I expected. But I guess we have to accept that. I think there shouldn't be any side effect to using this lib, like clearing all of the local storage. The view may want to use the local storage too.

gmarty: Also, the `Object.freeze()` should be recursive.

app/js/lib/foxbox/settings.js
gmarty: Okeydoke!

gmarty: https://github.com/fxbox/app/issues/143

app/js/lib/foxbox/settings.js
azasypkin: Okay, then will use this approach.

azasypkin:
@gmarty r?

Second commit is ready for review (for the first one I've asked r? in a separate PR).

Thanks!

app/js/lib/foxbox/services.js
```diff
@@ -203,6 +176,35 @@ export default class Services extends EventDispatcher {
...
+   *
+   * @return {Promise<Map<string, BaseService>>}
+   * @private
+   */
+  [p.getCache]() {
+    if (this[p.cache]) {
+      return this[p.cache];
+    }
+
+    return this[p.cache] = this[p.db].getServices()
```
azasypkin: Here was a bug, ```[p.cache]``` should have been a promise with ```Map``` value, not just ```Map``` in this case ```getCache``` can be called simultaneously from several places and all calls will wait on the single promise. Before that it could happen that view does two requests, first creates empty map and waits for the result to populate it with, but second request thinks that map is ready (but it's still empty).

azasypkin:
I've rebased PR on the latest master, so it's ready for review whenever you have time :)

azasypkin:
Gentle ping @gmarty :)

tests/unit/lib/foxbox/api_test.js
```diff
@@ -2,7 +2,7 @@ import { waitForNextMacroTask } from '../../test-utils';
 import API from 'js/lib/foxbox/api';
 
 describe('API >', function () {
```
gmarty: It'd be nice to start adding the annotations for tests when we modify unit tests files (`/** @test {API} */`).
The documentation generator will land soon.

gmarty:
I'm so sorry about my late review...
It looks all good to me. Is there anything I need to update or to check with the Philips Hue class?

r+

tests/unit/lib/foxbox/api_test.js
azasypkin: Right!

azasypkin:
> I'm so sorry about my late review...

Np, thanks for review!

> Is there anything I need to update or to check with the Philips Hue class?

I think you may want to add watchers for light statuses, but theoretically everything should work as it worked before :)

PR #142, opened 2016-05-03
Run localhost over HTTPS (fixes #131). r=gmarty

### LINKED ISSUE #131: Run localhost over HTTPS
Now that we have HTTPS connection end to end we need to run localhost over HTTPS for both the development and testing phases.

The final goal is to disallow all the non secure connections:
* Redirect the app from HTTP to HTTPS
* Redirect the registration server similarly
* Disallow non HTTPS connection to the box

azasypkin:


azasypkin:
Self signed certificate is valid for 2 years, when you run browser for the first time you should add exception for it.

gmarty:
r+!
The code looks good to me but I'd like to see some documentation. Can you add a few words in the `How to build?` section of the main README about adding an exception?
Also what happens after 2 years? This may need documenting too.

tests/foxbox-simulator/http-server.js
```diff
@@ -13,4 +15,10 @@ foxboxSimulator.get('/ping', (req, res) => {
   res.status(204).end();
 });
 
-foxboxSimulator.listen(3000);
+https.createServer(
+  {
+    key: fs.readFileSync(path.join(__dirname, '../../certs/key.pem')),
+    cert: fs.readFileSync(path.join(__dirname, '../../certs/cert.pem')),
```
gmarty: Why are the tests linted differently from the rest of the code? The `.eslintrc` file in `test/` should extend the one at the root level.

PR #145, opened 2016-05-05
Add a command to generate documentation. r=azasypkin

gmarty:
This patch adds a command to generate documentation from the source code.
Open questions:
* Shall we host the generated HTML files somewhere? If so where?
* Shall we commit the generated files? Possibly via a precommit hook.
* The optional parameters syntax used by Closure Compiler is not supported by esdoc. Shall we write an esdoc plug-in for that?

What do you think @azasypkin?

### DIFF
--- a/esdoc.json
+++ b/esdoc.json
@@ -1,0 +1,9 @@
+{
+  "source": "./app/js/lib/foxbox",
+  "destination": "./doc",
+  "title": "Project Link web app",
+  "test": {
+    "type": "mocha",
+    "source": "./tests"
+  }
+}
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "scripts": {
     "build": "gulp build",
     "test": "gulp test",
+    "doc": "rm -rf doc && esdoc -c esdoc.json",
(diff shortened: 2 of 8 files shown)


azasypkin:
Looks good to me r+!

> Shall we host the generated HTML files somewhere? If so where?

Yes, but maybe not now. I mean it should not be a part of deployed app, but rather a separate deployment. In my head once foxbox lib is in a separate repo, its own gh-pages will be the right place. Alternative solution would be to deploy docs to fxbox/app gh-pages, but deploy app to a separate repo (eg. fxbox/app-staging, fxbox/app-nightly, fxbox/app-demo-2016-04-29 :)). 

 
> Shall we commit the generated files? Possibly via a precommit hook.

I have no opinion on this, but it seems we don't need it right now, if developer wants to see docs it's easy enough to generate the latest docs locally. So I'd rather include ```docs``` folder into ```.gitignore```. If foxbox lib was in a separate repo then yes, precommit hook is a good choice. 

> The optional parameters syntax used by Closure Compiler is not supported by esdoc. Shall we write an esdoc plug-in for that?

You know I'm not fully convinced about using Google Closure Compiler syntax :) But anyway I'm totally fine with any of these 3 choices:
* Just live with it for now, doesn't sound like something very critical;
* Write esdoc plugin if it's easy enough to do;
* Switch to standard jsdoc optional parameter notation _if esdoc supports it_. Also wondering if the latest closure compiler recognizes JSDoc standard optional parameter notation.

PR #148, opened 2016-05-09
Migrate foxbox/settings private fields to new symbol-based naming system. Settings should listen for Storage event. Emit dedicated event per settings item change (fixes #130, fixes #143, fixes #144). r=gmarty

### LINKED ISSUE #130: [Refactoring] Migrate foxbox/settings private fields to new symbol-based naming system

### LINKED ISSUE #143: Emit one change event per item in settings
As discussed in https://github.com/fxbox/app/pull/141/files, the settings class should emit an event when any of its item changes (at the moment we only have `session` and `polling-setting`).

### LINKED ISSUE #144: Settings class must listen to storage event
Whenever a local storage value is edited outside of the lib (by a 3rd party script or using the devtools), the state may get out of sync and lead to unpredictable effect.
We should listen to the `storage` event and update the internal value in Settings accordingly. This will allow us to edit some settings in the devtools and can be useful for demos on mobile.

azasypkin:


tests/unit/lib/foxbox/settings_test.js
```diff
@@ -0,0 +1,214 @@
+import Settings from 'js/lib/foxbox/settings';
```
azasypkin: Just a few the most important tests to start with.

azasypkin:
r? @gmarty 

What do you think?

app/js/lib/foxbox/settings.js
```diff
@@ -22,178 +21,203 @@ const API_VERSION = 1;
...
+    window.addEventListener('storage', this[p.onStorage].bind(this));
 
-      resolve();
-    });
+    Object.seal(this);
   }
 
-  get configured() {
-    return this._configured;
+  get [settings.CONFIGURED.key]() {
```
gmarty: I'm a bit concerned that this syntax may not work well with the doc generator. We won't get the getter/setter in the documentation.
What do you think?

app/js/lib/foxbox/settings.js
```diff
@@ -203,4 +227,124 @@ export default class Settings extends EventDispatcher {
...
+    if (currentValue === newValue) {
+      return;
+    }
+
+    this[p.values].set(setting, newValue);
+
+    if (newValue !== this[p.getDefaultSettingValue](setting)) {
+      this[p.storage].setItem(`${PREFIX}${setting.key}`, newValue);
+    } else {
+      this[p.storage].removeItem(`${PREFIX}${setting.key}`);
```
gmarty: :+1: Nice that using the local storage is avoided on default value.

gmarty:
I like it a lot. Sounds much more efficient than the previous code. I have a concern about doc generator, but I'd understand if there's no practical workaround to it, so r+!

I took a look at some common lib wrapper around localStorage looking for potential performance issues, but it seems that localStorage is fast enough nowadays to not worry about it.

app/js/lib/foxbox/settings.js
azasypkin: Yep, true, here is what I see:
screenshot from 2016-05-11 14-14-13

So yeah, maybe let's just not use computed property names and use plain old names for getters and setters. It's small duplication, but benefits outweigh (docs and just much easier to read).

What do you think?

gmarty: Yes, I like the benefit of having a clean doc generated automatically. Let's use static name for the getters and setters.

azasypkin:
> I took a look at some common lib wrapper around localStorage looking for potential performance issues, but it seems that localStorage is fast enough nowadays to not worry about it.

Yeah, in the past days reading from localStorage was a disk IO operation, nowadays it's done in memory most of the time, so we should be fine I think.

app/js/lib/foxbox/settings.js
azasypkin: Done! waiting for travis.

PR #150, opened 2016-05-13
Re-enable tags management based on the taxonomy instead of IndexedDB (fixes #90, fixes #98). r=gmarty

### LINKED ISSUE #90: There is no way to know which light you are adding to a recipe
Please see below screenshot:
captura de pantalla 2016-04-15 a las 12 05 20

Should it be expected to have a number or name to differenciate them? or should it be the tag (which btw is not saved if it is selected for a service)?

### LINKED ISSUE #98: When we tag service, its getters and setters should have the same tag as well
We need this for Recipe page, so that we can distinguish getters and setters there.

azasypkin:


app/css/icons/tag.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="30" width="30" version="1.1"><g transform="matrix(.019697 0 0 -.019697 .079568 27.813)"><path fill="#fff" d="m448 1088q0 53-37.5 90.5t-90.5 37.5-90.5-37.5-37.5-90.5 37.5-90.5 90.5-37.5 90.5 37.5 37.5 90.5zm1067-576q0-53-37-90l-491-492q-39-37-91-37-53 0-90 37l-715 716q-38 37-64.5 101t-26.5 117v416q0 52 38 90t90 38h416q53 0 117-26.5t102-64.5l715-714q37-39 37-91z"/></g></svg>
```
gmarty: This file can be minified further. I usually only use a `viewBox` attribute in lieu of `height` and `width`. The `version` attribute can be removed to.
It's better if the `transform` tag is removed and its effect merged into the `path` values.

I can do that operation for you if you want.

app/js/views/service-tags.jsx
```diff
@@ -0,0 +1,98 @@
...
+    if (!tag || !tag.trim()) {
+      return;
+    }
+
+    service.addTag(tag.trim())
+      .catch((err) => {
+        console.error('Could not add a new tag: %o', err);
+      })
+      .then(() => {
+        this.setState({ tags: service.tags });
```
gmarty: This is better to do optimist updates.
You set the state immediately before doing the request to the box and if anything go wrong, then you undo the change.
This gives suprisingly a fast feeling to the UI.

app/js/views/service-tags.jsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  onAddTag() {
+    const service = this.state.service;
+    if (!service) {
+      return;
+    }
+
+    const tag = prompt('Enter new tag name');
+    if (!tag || !tag.trim()) {
```
gmarty: nit: can you avoid calling `trim()` twice?

app/js/views/service-tags.jsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  onRemoveTag(tag) {
+    const service = this.state.service;
+    service.removeTag(tag)
+      .catch((err) => {
+        console.error('Could not remove tag: %o', err);
+      })
+      .then(() => {
+        this.setState({ tags: service.tags });
```
gmarty: Same here, apply optimistic UI.

app/css/icons/tag.svg
azasypkin: Ah, you can teach me :) How would you merge ```transform``` directly to the ```path``` (without manually doing matrix math)?

azasypkin: Ah nevermind, I've just noticed ```Path -> Simplify``` in Inkscape that did the job :)

app/js/views/service-tags.jsx
azasypkin: Agree, this piece of code is `under construction` :) Trying to figure out how to do it better!

app/css/icons/tag.svg
gmarty: Yes, Inkscape is your friend! I don't usually blindy trust svgo. In simple cases like this one, I make sure that there is only one path defined and no useless tags.

app/js/views/service-tags.jsx
azasypkin: Yep I can do something like this to guard against ```null``` and call ```trim``` only once:
```
const tag = (prompt('Enter new tag name') || '').trim();
```

app/js/views/tag-list.jsx
```diff
@@ -1,25 +1,17 @@
...
-          foxbox={this.foxbox}/>
-      )
-    );
+    let tagNodes = this.props.tags.map((tag) => {
+      return (<li key={tag} className="tag-list__item">
+        {tag}
+        <button className="tag-list__item-remove"
+                type="button"
+                onClick={this.props.onRemoveTag.bind(null, tag)}>
+          ✕
```
gmarty: nit: The typo geek in me says that this is the sign for multiplication, not a ballot X. You can try one of these instead: ✗✘⮽, but I'd rather use a SVG image (maybe the plus.svg rotated by 45deg or a minus sign).

azasypkin: > The typo geek in me says that this is the sign for multiplication, not a ballot X

Nope, iirc it's unicode ```cross mark``` [1] :)

> maybe the plus.svg rotated by 45deg

As far as I remember it's too fat and white, but let me check it out.

[1] http://www.fileformat.info/info/unicode/char/274c/index.htm

app/css/views/service.css
```diff
@@ -0,0 +1,6 @@
+.service__edit-tags-link {
+  width: 3rem;
+  height: 3rem;
+  padding-left: 1rem;
+}
```
gmarty: We should really use flexbox on the header. When the title is too big or the window width too narrow, the tag icon is wrapped horribly. Instead, the title should be ellipsis'ed and the icon should still be completely visible with nice padding around.

azasypkin: Right, I just use existing approach from recipe view :p We can improve header layout in a separate bug for all views, how does it sound?

app/js/controllers/service-tags.js
```diff
@@ -0,0 +1,14 @@
...
+import ServiceTagsView from 'js/views/service-tags';
+
+export default class ServiceTagsController extends Controller {
+  main(id) {
+    ReactDOM.render(React.createElement(ServiceTagsView, {
+      id,
+      foxbox: this.foxbox,
+    }), this.mountNode);
+  }
+}
```
gmarty: Many controllers look the same with only the React tag changing.
I wonder if we can do something clever and avoid code repetition. Maybe using a kind of factory.
Anyway, that's not needed for this PR.

azasypkin: Yep, I have the same feeling, we need to workout better way...

gmarty:
This approach works for me! Feel free to ping me for a proper r? when you're ready.

app/js/lib/foxbox/recipes.js
```diff
@@ -193,6 +193,7 @@ export default class Recipes {
           id: getter.id,
           kind: getter.kind,
           name: name || getter.adapter,
+          tags: getter.tags,
```
azasypkin: If this tag idea sticks, we can also use tags inside recipe name/label/text.

app/js/lib/foxbox/services/base.js
```diff
@@ -151,6 +153,80 @@ export default class BaseService extends EventDispatcher {
   }
 
   /**
+   * Returns list of service tags.
+   *
+   * @return {Array<string>}
+   */
+  getTags() {
```
azasypkin: Made it a method instead of property, to make tag management less confusing (when you have an array property it may seem that just pushing/popping an element to this list will add/remove tag).

azasypkin:
>  I'm still trying to figure out why login tests started to fail.... but this should not affect PR anyway.

Okay, found and fixed - in ```service_test.js``` React methods were stubbed without using sinon sandbox that affected react for other tests as well, using proper sinon sandbox fixed the issue. I'm still not sure why my PR made it perma red instead of intermittent....

app/css/icons/remove.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 30"><path fill="#000" d="M 6.6191408,1.4609377 C 4.8997398,3.1803387 3.1803387,4.8997398 1.4609377,6.6191408 4.2545575,9.4127605 7.0481774,12.20638 9.8417972,15 7.0481774,17.79362 4.2545575,20.587239 1.4609377,23.380859 3.1803387,25.10026 4.8997398,26.819661 6.6191408,28.539062 9.4127605,25.745442 12.20638,22.951823 15,20.158203 c 2.79362,2.79362 5.587239,5.587239 8.380859,8.380859 1.719401,-1.719401 3.438802,-3.438802 5.158203,-5.158203 C 25.745442,20.587239 22.951823,17.79362 20.158203,15 22.951823,12.20638 25.745442,9.4127605 28.539062,6.6191408 26.819661,4.8997398 25.10026,3.1803387 23.380859,1.4609377 20.587239,4.2545575 17.79362,7.0481774 15,9.8417972 12.20638,7.0481774 9.4127605,4.2545575 6.6191408,1.4609377 Z" /></svg>
```
gmarty: Have you used svgo on these icons? It looks like they can be compressed even further.

app/js/controllers/service-tags.js
gmarty: Let's hold on this discussion for now until we find a better framework for the app.

app/css/views/service.css
gmarty: That works for me. Please open an issue in case we forget.

app/js/lib/foxbox/services/base.js
```diff
@@ -151,6 +153,80 @@ export default class BaseService extends EventDispatcher {
...
+    // instance (eg. recipes view).
+    const servicesSelector = { services: { id: this[p.id] }, tags: tag };
+    const channelsSelector = { channels: { service: this[p.id] }, tags: tag };
+
+    return Promise.all([
+      this[p.api].post('services/tags', servicesSelector),
+      this[p.api].post('channels/tags', channelsSelector),
+    ])
+    .catch((error) => {
+      this[p.tags].delete(tag);
```
gmarty: Yay! Optimistic UI by default! :+1:

app/js/views/tag-list.jsx
```diff
@@ -1,25 +1,16 @@
...
-          key={tag.id}
-          id={tag.id}
-          name={tag.name}
-          checked={tag.checked}
-          serviceId={this.props.serviceId}
-          foxbox={this.foxbox}/>
-      )
-    );
+    let tagNodes = this.props.tags.map((tag) => {
+      return (<li key={tag} className="tag-list__item">
```
gmarty: nit: what about using the short notation for arrow functions with a return here?

app/js/views/services-list-item.jsx
```diff
@@ -306,6 +314,15 @@ export default class ServicesListItem extends React.Component {
         return this.renderGenericService();
     }
   }
+
+  getServiceName() {
+    // If service has tags, let's use them as service identifier.
+    const nameFromTags = this.service.getTags().reduce((name, tag) => {
+      return name ? `${name}, ${tag}` : tag;
+    }, '');
```
gmarty: Is this not enough?
```js
const nameFromTags = this.service.getTags().join(', ')
```

app/js/views/themes-new.jsx
```diff
@@ -269,6 +269,15 @@ export default class ThemesNew extends BaseView {
       </select>
     );
   }
+
+  getChannelName(channel) {
+    // If channel has tags, let's use them as channel identifier.
+    const tags = channel.tags.reduce((tags, tag) => {
+      return tags ? `${tags}, ${tag}` : tag;
+    }, '');
```
gmarty: see comment above.

gmarty:
It looks good to me. I left a few questions/nits.
Also use svgo on the icons and you'll be good to land. r+

app/css/icons/remove.svg
azasypkin: Nope :) Didn't know about `svgo` btw, thanks!

azasypkin: Wow, ```0.774 KiB - 73.1% = 0.208 KiB``` \0/

app/css/views/service.css
azasypkin: Done https://github.com/fxbox/app/issues/156

app/js/views/services-list-item.jsx
azasypkin: Damn, I can't explain why I used this super-complex-and-clumsy solution :D Will definitely remove it!

app/js/views/tag-list.jsx
azasypkin: Yep, why not.

app/js/views/themes-new.jsx
azasypkin: Fixed, thanks!

app/js/views/services/default.jsx
```diff
@@ -1,25 +1,20 @@
...
-      tags: [],
-    };
-
     this.foxbox = props.foxbox;
     this.service = props.service;
 
+    this.state = {
+      name: this.service.name,
+    };
+
```
gmarty: nit: moving this block kind of pollutes the git diff.

app/js/views/services/light.jsx
```diff
@@ -1,25 +1,20 @@
...
-      tags: [],
-    };
-
     this.foxbox = props.foxbox;
     this.service = props.service;
 
+    this.state = {
+      name: this.service.name,
+    };
+
```
gmarty: nit: same here.

PR #152, opened 2016-05-24
Use viewBox attribute in SVG. r=azasypkin

gmarty:


azasypkin:
Yeah, looks good to me, r=me

PR #153, opened 2016-05-24
Compatibility with iOS. r=azasypkin

gmarty:
Adding support for Safari on iOS 9.3.1:
* Transpile es2015 features missing (`transform-es2015-classes` is used because computed prototype methods are not supported, but ideally we should only transpile such methods)
* Add polyfills for `fetch` and `URLSearchParams`
* Use `URLSearchParams` as a standalone class, not as a property of an instance of a `URL` object

I haven't noticed anything wrong with the app but we should probably run the test suites on a device too.

@azasypkin Can you review it when you have time?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,18 @@
 {
-  "presets": ["react"],
-  "plugins": ["transform-es2015-modules-amd"]
+  "presets": [
+    "react"
+  ],
+  "plugins": [
+    "transform-es2015-modules-amd",
+    "check-es2015-constants",
+    "transform-es2015-block-scoping",
+    "transform-es2015-arrow-functions",
+    "transform-es2015-parameters",
+    [
+      "transform-es2015-classes",
+      {
+        "loose": true
+      }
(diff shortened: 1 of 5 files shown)


.babelrc
```diff
@@ -1,4 +1,18 @@
 {
-  "presets": ["react"],
-  "plugins": ["transform-es2015-modules-amd"]
+  "presets": [
+    "react"
+  ],
+  "plugins": [
+    "transform-es2015-modules-amd",
+    "check-es2015-constants",
```
azasypkin: __issue:__ looks like you forgot to include these plugins with the proper version to ```package.json```

azasypkin:
Looks good to me (don't have iOS to check though), just add packages for all new plugins to `package.json` with the proper version and in sorted order :)

r+, thanks!

.babelrc
gmarty: I added them so that it's still compatible with npm 2.

PR #155, opened 2016-05-25
Replace deprecated ChannelKind with Feature. r=gmarty

azasypkin:


app/js/lib/foxbox/recipes.js
```diff
@@ -48,40 +48,31 @@ export default class Recipes {
...
 
-          switch (channel.kind) {
-            case 'ThinkerbellRuleOn':
-              if (channel.supports_fetch) {
-                rule.getEnabled = channelId;
-              } else if (channel.supports_send) {
-                rule.setEnabled = channelId;
-              }
+          switch (channel.feature) {
+            case 'thinkerbell/is-rule-enabled':
```
azasypkin: Now single channel can be both setter and getter \0/

app/js/lib/foxbox/recipes.js
```diff
@@ -201,30 +193,26 @@ export default class Recipes {
   }
 
   getSetters() {
-    // Currently we support only TTS, camera and connected lights as actuators.
-    const supportedKinds = [
-      {
```
azasypkin: No more clumsy Extension Channel Kind!

app/js/lib/foxbox/services/base.js
```diff
@@ -290,12 +291,22 @@ export default class BaseService extends EventDispatcher {
...
+   * supports fetch operation.
    * @private
    */
   [p.getFetchChannel](selector) {
-    return this[p.getChannels](selector).find(
+    const channel = this[p.getChannels](selector).find(
       (channel) => channel.supports_fetch
     );
+
+    if (!channel) {
```
azasypkin: I found these error messages useful when integrating new watchers/getter/setters for services.

app/js/lib/foxbox/services/base.js
```diff
@@ -304,39 +315,21 @@ export default class BaseService extends EventDispatcher {
...
-
-    switch (operationKind) {
-      case 'TakeSnapshot':
-        return 'Unit';
-      case 'LightOn':
-        return 'OnOff';
-      default:
-        return operationKind;
-    }
+    return channel;
```
azasypkin: I'll need your help to verify that this still works for Lights

app/js/lib/foxbox/services/light.js
```diff
@@ -15,11 +15,13 @@ export default class LightService extends BaseService {
   }
 
   isAvailable() {
-    return this.get('available').then((response) => response.OnOff === 'On');
+    return this.fetch('device/available')
+      .then((response) => response.OnOff === 'On');
```
azasypkin: With new ```supports_fetch``` signature we can eventually simplify this even more to smth like ```.then((response) => response === 'On');```, but let's wait for the foxbox to stabilize this time.

azasypkin:
r? @gmarty 

Should be ready for review so that we can work again with the foxbox master :)

app/js/lib/foxbox/services/base.js
```diff
@@ -232,7 +237,7 @@ export default class BaseService extends EventDispatcher {
    * super.teardown() method as well.
    */
   teardown() {
-    for (let watcher of this[p.watchers].values()) {
+    for (const watcher of this[p.watchers].values()) {
```
gmarty: Does it work natively on Firefox? I think we're now transpiling `const` and `let` so that's not really an issue.

azasypkin: Right, Fx doesn't support it yet, I've made it const because of transpiling andeslint that was complaining. So if we turn off transpiling later and Fx still doesn't support it we'll have to change it back :/

app/js/lib/foxbox/services/base.js
```diff
@@ -65,40 +64,46 @@ export default class BaseService extends EventDispatcher {
...
    */
-  set(selector, value = '') {
-    const setter = this[p.getSendChannel](selector);
-    const setterType = this[p.getSetterValueType](setter.kind);
-
-    return this[p.api].put(
-      'channels/set',
-      [[{ id: setter.id }, { [setterType]: value }]]
-    );
+  send(selector, value = null) {
```
gmarty: Why renaming to `send` and `fetch`? `get` and `set` are pretty standard. Is it to match the underlying logic of the box?

app/js/lib/foxbox/services/base.js
azasypkin: hmm, now when I read this code, I'm not sure how it should work in reality :) Do you think it should be `let` and eslint should not complain instead?

app/js/lib/foxbox/services/base.js
azasypkin: > Is it to match the underlying logic of the box?

Yep, to match their public API + now we officially call getters/setters as channels, so I thought send and fetch works better with channels.

What do you think?

app/js/lib/foxbox/services/base.js
gmarty: I think it is a valid use for `const`.
Now we can make this eslint rule an error rather than a warning.

app/js/lib/foxbox/services/base.js
gmarty: This library abstracts the communication with the box, it doesn't have to match perfectly the taxonomy. I like `set` and `get` better as there are easy to understand and to remember. I don't think they're incompatible with the term `channels`.
Anyway, I leave the final word to you if you think that's better to rename them.

gmarty:
I left a comment. It looks good and I checked on a set of Hue.
r+!

app/js/lib/foxbox/services/base.js
azasypkin: > I think it is a valid use for const.

Right, just double checked with [1] :)

> Now we can make this eslint rule an error rather than a warning.

Good point, I'll make eslint to consider it as an error

[1] https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/for...of

app/js/lib/foxbox/services/base.js
azasypkin: I'll revert it back to ```get``` and ```set``` then, since there is a rule - `if not critical, don't change if there is no 100% agreement` :)

azasypkin:
> I left a comment. It looks good and I checked on a set of Hue.
r+!

Nice, thanks! Handling your comments...

Issue #46, opened 2016-04-01
Optimise React for production builds


gmarty:
Babel react conserves the `propTypes` of components (which is both expected and wanted).
But in production mode they add no benefits to the components so we want to remove them.

Let's enquire and see if it's a common practice in the React world and if there are already tools to do that.

gmarty:
It turns out there are a couple more optimisations we can do to make more efficient production build of React:
https://medium.com/doctolib-engineering/improve-react-performance-with-babel-16f1becfaa25

gmarty:
We also need to use the `.min.js` versions of React scripts for better performance in production.

PR #163, opened 2016-06-03
Concatenate and minify the generate scripts. r=gmarty

azasypkin:


.babelrc
```diff
@@ -3,11 +3,12 @@
     "react"
   ],
   "plugins": [
-    "transform-es2015-modules-amd",
```
azasypkin: Now it's done by Rollup (but I still use this plugin when compile app for unit tests)

.babelrc
```diff
@@ -3,11 +3,12 @@
...
     "check-es2015-constants",
     "transform-es2015-block-scoping",
     "transform-es2015-arrow-functions",
     "transform-es2015-parameters",
+    "transform-es2015-destructuring",
```
azasypkin: Uglify harmony branch doesn't support destructuring yet, as well as Edge (this thing doesn't support a ton of stuff, soon I'll try to make it work), so I'd rather transpile this feature as well, what do you think?

app/components/mvc.js
```diff
@@ -1 +0,0 @@
-../../node_modules/fxos-mvc/dist/mvc.js
```
azasypkin: That the thing I wanted to do for a long time, now we copy these external components at the build stage.

app/js/bootstrap.js
```diff
@@ -2,11 +2,19 @@
...
-  map: {
-    '*': {
-      'react': 'components/react',
-    },
-  },
+  map: { '*': { 'react': 'components/react' } },
 });
 
-require(['js/app.js']);
+const polyfills = [];
```
azasypkin: Idea is to load polyfills only when we need them (and Edge will definitely require more polyfills), what do you think?

gulpfile.js
```diff
@@ -39,80 +47,130 @@ gulp.task('lint', function() {
...
-    .pipe(gulp.dest(DIST_APP_ROOT));
+gulp.task('compile-app', function() {
+  return gulp
+    .src(`${APP_ROOT}js/*.js`)
+    .pipe(
+      rollup({
+        sourceMap: true,
+        format: 'amd',
+        external: [
+          'components/mvc',
```
azasypkin: This `external` property not strictly necessary, I just wanted to reduce number of warnings generated by rollup

gulpfile.js
```diff
@@ -39,80 +47,130 @@ gulp.task('lint', function() {
...
-      `${APP_ROOT}**`,
-      `!${APP_ROOT}js/**/*.js`, // do not copy js
-    ])
-    .pipe(gulp.dest(DIST_APP_ROOT));
+gulp.task('compile-app', function() {
+  return gulp
+    .src(`${APP_ROOT}js/*.js`)
+    .pipe(
+      rollup({
+        sourceMap: true,
```
azasypkin: Now we have es6 sourcemaps! That means that even debug build will have single app.js, but source maps should work well for us - I've played with them for a few days and haven't noticed any troubles. If in the future we encounter any problems, we can disable Rollup for debug build (like we do for unit tests).

gulpfile.js
```diff
@@ -39,80 +47,130 @@ gulp.task('lint', function() {
...
-        })
-      )
-      .pipe(gulp.dest(`${DIST_TESTS_ROOT}unit/`));
-  } catch (e) {
-    console.error('Error occurred while process unit test files with Babel', e);
-  }
+gulp.task('compress-external-modules', function() {
+  return gulp
+    .src([
+      `${DIST_APP_ROOT}js/alameda.js`,
```
azasypkin: Actually these files were pretty big and don't have min versions :/

gulpfile.js
```diff
@@ -39,80 +47,130 @@ gulp.task('lint', function() {
...
-    `${DIST_APP_ROOT}**/*.md`,
-  ]);
+/**
+ * Applies Babel transform to the app/tests JS and JSX files. The only
+ * difference between normal app build and build for unit tests is that Rollup
+ * is not used here.
+ */
+gulp.task('compile-unit-tests', function() {
+  return merge(
+    gulp
```
azasypkin: Here is the thing I don't really like, but I've struggled with Rollup & Unit tests too much already :) Ideally I wanted to run unit tests on the production build, but rollup doesn't export anything from the bundle...

package.json
```diff
@@ -19,38 +20,43 @@
   "license": "MPL2",
   "dependencies": {
     "alameda": "^1.0.0",
+    "fxos-mvc": "fxos-eng/mvc#master",
```
azasypkin: Moved all dev dependencies to the appropriate section. iirc dependencies section should contain only run time app dependencies.

azasypkin:
Hey @gmarty how does it look to you? I've deployed production version to http://azasypkin.github.io/app, so that you can try it.

.babelrc
gmarty: That sounds good. Just enabling destructuring makes it work on Edge or is there anything else obvious to do?

app/components/mvc.js
gmarty: :+1:

app/index.html
```diff
@@ -28,7 +28,7 @@
 <link rel="icon" href="img/icons/512.png" sizes="512x512">
 <link rel="icon" href="img/icon.svg" sizes="any" type="image/svg+xml">
 <title>Project Link</title>
-<script defer src="js/initapp.js"></script>
+<script src="js/alameda.js" data-main="js/bootstrap"></script>
```
gmarty: Doesn't it work with defer anymore?
I really wish we can split independent chunks of JavaScript with `defer` and `async`.

app/js/app.js
```diff
@@ -1,4 +1,4 @@
-import MainController from 'js/controllers/main';
+import MainController from './controllers/main';
```
gmarty: question: Why do you need to change the relative paths?

gulpfile.js
```diff
@@ -39,80 +47,130 @@ gulp.task('lint', function() {
...
+      // Don't copy JS, it will be compiled and copied on the compile step.
+      `!${APP_ROOT}js/**`,
+    ]),
+
+    // Module loader.
+    gulp.src('./node_modules/alameda/alameda.js')
+      .pipe(rename('js/alameda.js')),
+
+    // Components.
+    gulp.src('./node_modules/react/dist/react.min.js')
```
gmarty: YAY! `.min.js` files!

gulpfile.js
```diff
@@ -39,80 +47,130 @@ gulp.task('lint', function() {
...
+        ],
+      })
+    )
+    .pipe(sourcemaps.write('.'))
+    .pipe(gulp.dest(`${DIST_APP_ROOT}js`));
 });
 
 /**
- * Apply Babel transform to the JS and JSX files.
+ * Compiles app source for production i.e. processes source with Bable, Rollup
```
gmarty: typo: `Babel`.

.babelrc
azasypkin: Nope :/ Just tried, it requires few more things (at least I've noticed that Edge doesn't support "includes")

app/js/app.js
azasypkin: Two reasons:
* We use relative paths in the rest of the code - so for the sake of consistency;
* Rollup considers non-relative paths as external paths and doesn't include them into bundle (sounds reasonable)

app/index.html
azasypkin: Can you give me a little more details about `independent chunks` of js? I'm not sure I'm following...

About `defer` - yeah, I think I'll return `defer` here.

gulpfile.js
azasypkin: The funny thing that I don't see `.js.map` files for react... But anyway, I needed to debug react only once :)

gulpfile.js
```diff
@@ -39,80 +47,130 @@ gulp.task('lint', function() {
...
+    gulp.src('./node_modules/alameda/alameda.js')
+      .pipe(rename('js/alameda.js')),
+
+    // Components.
+    gulp.src('./node_modules/react/dist/react.min.js')
+      .pipe(rename('js/components/react.js')),
+    gulp.src('./node_modules/react-dom/dist/react-dom.min.js')
+      .pipe(rename('js/components/react-dom.js')),
+    gulp.src('./node_modules/fxos-mvc/dist/mvc.js')
+      .pipe(rename('js/components/mvc.js')),
```
gmarty: We still want to keep the non minified files for React in debug builds because they comes with additional checking and clever warning useful for debugging.

gulpfile.js
```diff
@@ -39,80 +47,130 @@ gulp.task('lint', function() {
...
+      rollup({
+        sourceMap: true,
+        format: 'amd',
+        external: [
+          'components/mvc',
+          'components/react',
+          'components/react-dom',
+        ],
+        plugins: [
+          rollupBabelPlugin({ presets: ['es2015-rollup'] }),
```
gmarty: Are you sure this preset does take the `.babelrc` file into account and is not transpiling all the es2015 features to es5?

gulpfile.js
azasypkin: Good point! I totally forgot about that, soo I can have `copy-app-production`, sounds good?

package.json
gmarty: Usually you list in devDep the dependencies needed to rebuild the project. It looks good like this but it's pretty useless because you can't run the code unless you build it. But that's fine I guess.

gulpfile.js
azasypkin: Heh, that maybe it, let me dig and check

package.json
azasypkin: Yeah, that's indented for npm modules that are distributed, but we have an app that's not distributed, hence the confusion :)

app/index.html
gmarty: If your app uses completely independent set of JavaScript files, you can bundle them separately (for instance, a main application code base and the analytics code) and load them with `defer` and `async`. This way it doesn't matter if one is loaded better or after the other because they're completely independent and their loading order is not relevant.

gulpfile.js
azasypkin: So that looks like a bug in plugin, adding additional settings should have complemented .babelrc, not entirely override it - I'll look how to fix that.

azasypkin: So that looks like a bug in plugin, adding additional settings should have complemented .babelrc, not entirely override it - I'll look how to fix that.

gulpfile.js
gmarty: Perfect!

gmarty:
That looks great! That's a step forward to a production ready app. We still need to optimise React (#46) and concatenate and minify the CSS (#158) but I'm glad we're tackling this issue.
r+ once all the comments are addressed.

app/index.html
azasypkin: > If your app uses completely independent set of JavaScript files, you can bundle them separately (for instance, a main application code base and the analytics code) and load them with defer and async. This way it doesn't matter if one is loaded better or after the other because they're completely independent and their loading order is not relevant.

Yeah, I understand, I meant what exactly do you want load separately in foxapp at the moment?

azasypkin:
Thanks for review, Travis is green and comments are handled!

PR #166, opened 2016-06-06
Optimise React for production builds (fixes #46). r=gmarty

### LINKED ISSUE #46: Optimise React for production builds

azasypkin:


app/js/views/services.jsx
```diff
@@ -69,7 +69,3 @@ export default class Services extends BaseView {
     );
   }
 }
-
-Services.propTypes = {
```
azasypkin: Here and below proptypes are the same as in the base class so no need to keep them.

Also plugin that removes propTypes has a small bug/missing point - it can't remove proptypes from the child classes (so propTypes for classes that inherit from BaseView and have their own propTypes won't be removed). We have only 3 of them now so it's not really an issue, but still. If I have time I'll investigate and maybe file issue.

package.json
```diff
@@ -48,7 +51,6 @@
     "express": "^4.13.4",
     "gulp": "^3.9.1",
     "gulp-babel": "^6.1.2",
-    "gulp-babel-external-helpers": "^2.0.2",
```
azasypkin: The less suspicious plugins the better :)

app/js/views/services.jsx
gmarty: Good catch. File an issue in their project and see what happens. If I understand correctly, it doesn't impact us right now.

.babelrc
```diff
@@ -16,5 +16,24 @@
...
+  "env": {
+    "development": {
+      "plugins": ["external-helpers"]
+    },
+
+    "production": {
+      "plugins": [
+        "external-helpers",
+        "transform-react-inline-elements",
+        "transform-react-constant-elements",
```
gmarty: I struggle to understand what's the benefit of this pass. If the generated code is minified, the constant should be inlined into the function if the minifier support inlining of invariants:
```js
// Before:
const Hr = () => {
  return <hr className="hr" />;
}
// After:
const _ref = <hr className="hr" />;
const Hr = () => {
  return _ref;
};
// After minification:
const Hr = () => {
  return <hr className="hr" />;
};
```

gmarty:
This patch looks great! Thanks.
r+

.babelrc
azasypkin: >  if the minifier support inlining of invariants:

Well it seems uglifyjs doesn't do this in the current setup, I remember that Google Closure compiler did this only in advanced mode.

But in our case I'd say we don't want to inline code related to jsx to keep as less memory allocations as possible. I can think of something like this:
```
// Code :
export default class Component extends React.Component {
  render() {
    const markup = <hr className="hr" />;
    return markup;
  }
}

// After Babel
var markup = React.createElement("hr", { className: "hr" });
var Component = function (_React$Component) {
  ........
  Component.prototype.render = function render() {
    return markup;
  };
  ........
}(React.Component);
``` 

So we share this immutable `markup` between instances. 

But №2 :) Just looked more closely and I don't see that `transform-react-constant-elements` changes anything in compiled version when `transform-react-inline-elements` is used... Didn't spend time to understand, I'd rather keep them both just in case, but can leave only `transform-react-inline-elements` if you want @gmarty ?

app/js/views/services.jsx
azasypkin: Yep, filed with reduced test case https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/issues/29

azasypkin: > If I understand correctly, it doesn't impact us right now.

It impacts, but it's very-very-very minor thing (just 3 noop proptypes are left inside minified file) :)

.babelrc
gmarty: I'm still very confuse about this inlining thing. For me, both examples are the same, they're invariants so it's just a matter of different syntax. Modern JIT inline invariant anyway so I don't really see the point.
Let's leave it as is for now if it's a good practice. Whatever it does or doesn't won't impact much the performance of the app.

azasypkin:
Thanks for review!

PR #168, opened 2016-06-07
Concatenate and minify the CSS resources + introduce CSS linter (fixes #158). r=gmarty

### LINKED ISSUE #158: Concatenate and minify the CSS resources

azasypkin:


.stylelintrc
```diff
@@ -0,0 +1,135 @@
+{
```
azasypkin: So here is just basic setup, there are several plugins that we can consider, but I'd rather consider and integrate them in a separate PR (bem-naming and combinator-selectors rules to name a few).

.stylelintrc
```diff
@@ -0,0 +1,135 @@
...
+    "color-named": "never",
+    "declaration-no-important": true,
+    "font-family-name-quotes": "always-where-recommended",
+    "font-weight-notation": "numeric",
+    "function-url-quotes": "always",
+    "selector-max-specificity": "0,2,0",
+    "selector-no-id": true,
+    "selector-root-no-composition": true,
+    "string-quotes": "single",
+    "declaration-block-properties-order": [
```
azasypkin: Mostly based on https://github.com/csscomb/csscomb.js/blob/master/config/yandex.json, but we can change if you think something looks very weird.

app/css/app.css
```diff
@@ -1,8 +1,23 @@
+@import 'FiraSans/font.css';
```
azasypkin: For now "compiled" app.css file will contain everything. Right now it's just 2.5kb gziped, so it sounds like we don't need to have several css bundles.

app/css/components/navigation-menu.css
```diff
@@ -11,40 +11,42 @@
...
 .navigation-menu__item-link {
   display: block;
 
   height: 4.75rem;
-  color: #fff;
   text-decoration: none;
-  font-size: 0;
-  background: url(../menu-icons/settings.svg) center 1.6rem no-repeat;
+  color: #fff;
+  background: url('../menu-icons/settings.svg') center 1.6rem no-repeat;
```
azasypkin: I chose single quote to not have mental switch between js and css (and we had more single quotes in css than double quotes :)). Let me know if you disagree.

app/js/views/user-logout-button.jsx
```diff
@@ -1,31 +0,0 @@
-import React from 'components/react';
```
azasypkin: Looks like we don't need this one anymore.

gulpfile.js
```diff
@@ -24,6 +24,11 @@ const rollupIncludePathsPlugin = require('rollup-plugin-includepaths');
 const rollupUglifyPlugin = require('rollup-plugin-uglify');
 const minifier = require('gulp-uglify/minifier');
 const uglifyjs = require('uglify-js');
+const postcss = require('gulp-postcss');
+const postcssImport = require('postcss-import');
+const postcssUrl = require('postcss-url');
+const cssnano = require('cssnano');
```
azasypkin: I'm experimenting with another css minifier at the moment, but it seems this one is actively developed, so likely I'm going to recomend cssnano :) At the moment we have relatively tiny css code base so I don't really care.

gulpfile.js
```diff
@@ -41,16 +46,21 @@ let registrationServerSimulator;
...
+    gulp
+      .src([`${APP_ROOT}**/*.{js,jsx}`, `${TESTS_ROOT}**/*.js`, './*.js'])
+      .pipe(eslint())
+      .pipe(eslint.format())
+      .pipe(eslint.failAfterError()),
+
+    gulp
+      .src('app/**/*.css')
+      .pipe(
+        stylelint({ reporters: [{ formatter: 'verbose', console: true }] })
```
azasypkin: Verbose mode includes pretty nice stat.. tell me if you find it too noisy.

azasypkin:
r?

Hey @gmarty how does it look to you? I've also deployed this PR to http://azasypkin.github.io/app

.stylelintrc
gmarty: I don't usually object things when it comes to CSS. I think most of the rules here make sense.
The only one I don't agree with is `color-named`. I like CSS colour names because it's easier for a human being to understand what is the colour. But I'm more curious about understanding the rationale than challenging this rule.

app/css/components/navigation-menu.css
gmarty: I agree! Single quotes are generally easier to read.

.stylelintrc
azasypkin: Yes, everything is arguable. Here are my pros:
* Not all colors have human readable format, so we'll definitely have both hex and human readable values at the same time. So we'll end up with color-inconsistency;
* When you edit images or design markup in photo editor - Gimp/Photoshop - you can easily copy color to/from there. You don't need to think what `magenta` or `teal` is; Or when you use color picker to grab color from another place, it always gives you hex or rgb - how could I know if it has human readable alternative?
* Looks like we have only 17 color names that are standartized [1], but what about the rest 123 from [2]? Does it mean that different browsers can choose whatever hex code for `CornflowerBlue` they want? This will be quite nasty bug if it's the case;
* Don't consider this one as pros, it's just about my personal feeling :) So for me using numbers everywhere gives me 100% confidence on the exact value of font-weight/color/border-width we're talking about.
So what are your good points of human readable colors?

[1] https://www.w3.org/TR/CSS22/syndata.html#value-def-color
[2] http://www.w3schools.com/colors/colors_names.asp

azasypkin: __Correction:__ about `only 17 color names that are standartized`, looks like there are so called `extended color codes` [1].

[1] https://www.w3.org/TR/2011/REC-css3-color-20110607/#svg-color

gmarty: As I said I don't have much opinion on that. Thanks for the explanation, it makes total sense to me.

azasypkin:
> r+! Thanks a lot!
Can you also please add a quick note about CSS standard on the CONTRIBUTING.md file?

Thanks a lot for review! Yep, will add a note to the CONTRIBUTING.md as well.

.stylelintrc
azasypkin: Good! If something goes wrong with it we can always get rid of this rule :)

PR #169, opened 2016-06-08
Port the deployment script to Gulp (fixes #132). r=gmarty

### LINKED ISSUE #132: Port the deployment script to Gulp
The current deployment script, `deploy.sh`, is executed until the end even if one of the previous steps fail, potentially leading to an unstable/buggy version of the app online.
If we use a Gulp task, the process will stop at the first error.

azasypkin:


gulpfile.js
```diff
@@ -416,3 +418,46 @@ gulp.task('test-e2e', function() {
     'stop-webserver'
   );
 });
+
+/**
+ * Deploys production-optimized app build to the "origin/gh-branch". Every new
```
azasypkin: So here is short description what it does. Honestly I like the fact that every deployment is a separate commit, so it's easy to track various revisions and get back to the last stable if needed. Also it's quite handy that not-stashed changes are deployed as well (but there will a red warning :)).

gulpfile.js
```diff
@@ -416,3 +418,46 @@ gulp.task('test-e2e', function() {
...
+gulp.task('deploy', ['build-production'], function() {
+  const wrapIntoPromise = (method, args) => new Promise((resolve, reject) => {
+    git[method].call(
+      git, args, (err, result) => err ? reject(err) : resolve(result)
+    );
+  });
+
+  return Promise
+    .all([
+      wrapIntoPromise('revParse', { args: '--short HEAD', quiet: true }),
```
azasypkin: This two git-related steps (git revision and warning about local changes) are not necessary, they bring a bit of complexity, but I found them useful. Anyway I'm not 100% convinced that super-duper needed, so let me know if you prefer to get rid of it.

azasypkin:
f? @gmarty

gulpfile.js
```diff
@@ -416,3 +418,46 @@ gulp.task('test-e2e', function() {
...
+      wrapIntoPromise('status', { args: '--porcelain', quiet: true }),
+    ])
+    .then((results) => {
+      const revision = results[0];
+      const status = results[1];
+      const deployDate = new Date().toUTCString();
+
+      let message = `Deployed on '${deployDate}', based on '${revision}'`;
```
azasypkin: I'm not good at labels :) But this is how it looks: https://github.com/azasypkin/app/commits/gh-pages

gmarty: Looks good. I don't think we need quotes around the date or the revision number.

gmarty: Actually not sure if the date is needed because commits are already stamped with the date and time.

gulpfile.js
```diff
@@ -416,3 +418,46 @@ gulp.task('test-e2e', function() {
...
+          '\x1b[31m',
+          `You have uncommitted changes that will be deployed!\n${status}`,
+          '\x1b[0m'
+        );
+
+        message += ' (includes uncommitted changes)';
+      }
+
+      return new Promise((resolve, reject) => {
+        merge(gulp.src(`${DIST_APP_ROOT}**/*`).pipe(ghPages({ message })))
```
azasypkin: Ah, forgot to mention - this `merge` around stream is workaround I was able to find so far, the issue is that `.on('end', ...` without `merge` wrapper is called at some weird moment and plugin doesn't have time to finish the job - it's definitely bug somewhere - just need more time to investigate.

gulpfile.js
azasypkin: Right, I thought about this too (plugin uses date in the commit message by default that's probably why I use it as well)... Right now I can't think of any use case for this date, I'll get rid of it, thanks!

gulpfile.js
gmarty: No I like this approach. It's good to be reminded if there are uncommitted files.

gulpfile.js
gmarty: I think it's better to keep the history too.

gmarty:
That looks great to me! r+
It's cool that you worked on this, I bumped into a deployment issue 2 days ago with npm3 and the bash script doesn't really fail.

PR #171, opened 2016-06-08
Get rid of fxos-mvc dependency. r=gmarty

azasypkin:


app/js/controllers/base.js
```diff
@@ -0,0 +1,9 @@
+export default class BaseController {
```
azasypkin: I've migrated only things we actually use to have less technical debt and bias as possible (so no teardown and "view.init()" method).

app/js/controllers/main.js
```diff
@@ -9,16 +8,23 @@ import DevController from './dev';
 
 import Foxbox from '../lib/foxbox/foxbox';
 
-export default class MainController extends RoutingController {
+const p = Object.freeze({
```
azasypkin: We used routing controller only in one place and we're going to migrate to some other framework, so here again I tried to migrate only things that we use and just merged RoutingController with MainController. If in the future we need it, it's pretty easy to extract it.

azasypkin:
r? @gmarty 

As always current production version of this PR is deployed here: http://azasypkin.github.io/app

app/js/controllers/main.js
gmarty: Have you considered moving the router to [react-router](https://github.com/reactjs/react-router)?
I think we should use it in the future, but this approach works for now.

azasypkin: > Have you considered moving the router to react-router?

Not yet

> I think we should use it in the future,

Yep, sounds good!

app/js/controllers/base.js
gmarty: I like this.

gmarty:
The patch looks good. I'm glad we're finally getting rid of this.
r+

azasypkin:
Thanks for review!

PR #172, opened 2016-06-09
Add support for the latest API. r=gmarty

azasypkin:


app/js/lib/foxbox/api.js
```diff
@@ -266,78 +266,78 @@ export default class API {
...
       } else {
-        valueChanged = newValue !== oldValue;
+        valueChanged = newValue !== channel.value;
       }
     }
 
     if (valueChanged) {
-      getter.value = Object.freeze(getterValue);
-      this[p.watchEventBus].emit(getter.id, getter.value);
+      // In ES6, a non-object argument will be treated as if it was a frozen
```
gmarty: I'm not sure I get this comment. Do you mean an empty object, e.g. `{}`?

azasypkin: No I mean, previously "channel.value" was always object (e.g. `{ OpenClosed: "Open"}`), but now it can be just primitive value (e.g. "Open"). In ES5 mode `Object.freeze("Open")` will be complaining, but in ES6 (like in all our supported browsers), everything will be ok :)

azasypkin: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze#Notes

app/js/lib/foxbox/api.js
```diff
@@ -9,16 +9,16 @@ const p = Object.freeze({
...
   // Private methods.
   getURL: Symbol('getURL'),
   onceOnline: Symbol('onceOnline'),
   onceAuthenticated: Symbol('onceAuthenticated'),
   onceDocumentVisible: Symbol('onceDocumentVisible'),
   onceReady: Symbol('onceReady'),
-  fetchGetterValues: Symbol('fetchGetterValues'),
-  updateGetterValue: Symbol('updateGetterValue'),
+  fetchChannelValues: Symbol('fetchChannelValues'),
+  updateChannelValue: Symbol('updateChannelValue'),
```
gmarty: I remember we used `get` and `set` in the service classes rather than `fetch` and `update`. Do you think we can use a consistent pattern all over the lib?

gmarty:
I'm not really familiar with the new taxonomy API but the code looks good to me. I left a couple of comments.
r+

app/js/lib/foxbox/api.js
azasypkin: Yep, I agree that we should have consistent patters, I can rename `fetchChannelValues` to `getChannelValues`, but `updateChannelValue` is quite different from BaseService.set. BaseService.set sets the value for the particular channel (and change is persisted in the foxbox/device), while `updateChannelValue` just updates channel value in internal app cache, so I'd prefer to keep it `update` here.

Sounds good?

gmarty: Yep, sounds good to me!

azasypkin:
Comment handled and travis is happy, merging :)

PR #175, opened 2016-06-10
Use arrow functions in gulpfile.js r=azasypkin

gmarty:


gmarty:
@azasypkin: How does it look to you?

gulpfile.js
```diff
@@ -282,37 +284,29 @@ gulp.task('webserver', function() {
...
-gulp.task('clobber-tests', function() {
-  return del(DIST_TESTS_ROOT);
-});
-
-gulp.task('clobber-doc', function() {
-  return del(DOC_ROOT);
-});
+gulp.task('clobber-app', () => del(DIST_APP_ROOT));
+gulp.task('clobber-tests', () => del(DIST_TESTS_ROOT));
+gulp.task('clobber-doc', () => del(DOC_ROOT));
```
gmarty: note: This is the only place where I used the return shorthand syntax.

gulpfile.js
```diff
@@ -85,14 +85,14 @@ gulp.task('copy-app-common', function() {
     gulp.src('./node_modules/whatwg-fetch/fetch.js')
       .pipe(rename('js/polyfills/fetch.js'))
   )
-  .pipe(gulp.dest(DIST_APP_ROOT));
+    .pipe(gulp.dest(DIST_APP_ROOT));
```
azasypkin: __question:__ Is it by accident or you think this alignment is better (here and in several places below)? Anyway up to you, I don't really care :)

gulpfile.js
azasypkin: Yeah, looks fine here

gulpfile.js
```diff
@@ -353,43 +347,43 @@ gulp.task('start-simulators', function() {
...
 });
 
-gulp.task('run-test-e2e', function() {
+gulp.task('run-test-e2e', () => {
   return gulp.src(`${TESTS_ROOT}{common,e2e}/**/*_test.js`, { read: false })
     .pipe(mocha());
 });
 
-gulp.task('test', ['clobber-tests', 'build-production'], function(cb) {
+gulp.task('test', ['clobber-tests', 'build-production'], (cb) => {
```
azasypkin: __proposal:__ it's becoming hard to track new warnings in EsLint output because of these 30 warnings related to `Unexpected block statement surrounding arrow body`. I'd rather disable this rule at all... What do you think?

azasypkin:
r=me, just one proposal and question.

gulpfile.js
gmarty: These alignements were not all consistent, so WebStorms code formatting put them all back together.

gulpfile.js
gmarty: Yeah that probably makes sense. Let's do it on another PR.

PR #176, opened 2016-06-13
Make app work in Microsoft Edge (fixes #173). r=gmarty

### LINKED ISSUE #173: Make app work in Microsoft Edge
At least at this stage I'd like to evaluate how much work it is to add Microsoft Edge support.

Will start with Vagrant VM [1].

[1] https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/linux/

azasypkin:


.babelrc
```diff
@@ -4,6 +4,7 @@
   ],
 
   "plugins": [
+    "array-includes",
```
azasypkin: Well, I've tried several ways to solve the issue:
* [core-js](https://www.npmjs.com/package/core-js) - it's a huge collection of different shims/polyfills, it's great when you need to shim a lot of stuff or you're not in a browser - bundle that it generates just for array.includes is unreasonably big :/
* [array-includes](https://www.npmjs.com/package/array-includes) - a bit better, but requires additional gulp task to compile browser-compatible bundle - didn't even try to write it, as it feels too much just for this shim :)
* And this one - just replace `includes` with babel plugin - easiest way and don't affect peformance in any way.

And the main factor is that Edge array.includes implementation is already at the [Preview stage](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/arrayprototypeincludeses2016)! So we can get rid of this plugin soon enough.

package.json
```diff
@@ -18,7 +18,7 @@
     "iot"
   ],
   "author": "",
-  "license": "MPL2",
+  "license": "MPL-2.0",
```
azasypkin: Finally decided to get rid of this annoying npm complaint about invalid license format :)

azasypkin:
@gmarty gentle ping :)

gmarty:
r+. Looks good to me.

PR #177, opened 2016-06-16
Polyfill for no IndexedDB support (fixes #165). r=gmarty

### LINKED ISSUE #165: Polyfill for no IndexedDB support
If IndexedDB is not supported by the browser, the app won't work at all. This impacts at least the following situations:
* Firefox in private mode
* iOS in web app (full screen app) mode (and webview probably), see #154

For such environments, we should use a in-memory storage instead of IndexedDB.

azasypkin:


azasypkin:
r? @gmarty 

How does it look to you? There are no unit tests here as I'm not sure if anyone needs this app in the future anyway :) App is deployed to my gh-pages.

app/js/lib/foxbox/common/in-memory-storage.js
```diff
@@ -0,0 +1,100 @@
+const p = Object.freeze({ db: Symbol('db') });
```
gmarty: Can you start with `'use strict';` for the sake of consistency?

app/js/lib/foxbox/db.js
```diff
@@ -25,31 +18,44 @@ const DB_VERSION = 1;
 
 const DB_SERVICE_STORE = 'services';
 
+class FoxboxIndexedDbStorage extends IndexedDbStorage {
```
gmarty: I'd rather call it something like `FoxBoxDbStorage` as it isn't always backed by IndexedDB.
What do you think?

app/js/lib/foxbox/common/indexed-db-storage.js
```diff
@@ -0,0 +1,159 @@
+const p = Object.freeze({
```
gmarty: nit: same here.

app/js/lib/foxbox/db.js
gmarty: Got it. Please ignore my comment.

gmarty:
It looks good to me. r+
Excellent job!

app/js/lib/foxbox/common/in-memory-storage.js
azasypkin: Ah, yeah, sure

azasypkin:
Thanks for review! Nits fixed, checked the case when IndexedDB is null (just to be 100% sure) and waiting for travis.

PR #178, opened 2016-06-21
Fix app<->foxbox breaking changes. r=gmarty

azasypkin:
App deployed to my gh-pages. To be used with the latest foxbox.


gmarty:
The code looks good to me. r+

